PR #9, opened 2026-01-28
Fix MCP request ID collisions

DeanoC:
Fixes a bug where concurrent MCP requests could share the same request ID when issued within the same millisecond, causing tools/resources to hang or return empty. Adds a per-transport counter suffix to guarantee unique IDs.


agnt-gg:
Thank you @DeanoC !!

PR #31, opened 2026-04-27
feat(chutes): add E2EE provider support

alex-drocks:
Adds a first-class chutes.ai provider that uses the end-to-end encrypted fetch transport.

Includes:
- ChutesE2EECrypto.js   – key agreement + AES-GCM symmetric crypto
- ChutesE2EEFetchTransport.js – encrypted fetch layer with handshake
- ChutesDiscoveryManager.js – model discovery via /supported-models API
- Frontend icons, provider config, and store wiring
- Required dependencies: mlkem, @stablelib/chacha20poly1305

Chutes provider in options:
image

Chutes provider confirmed API key:
image

Chutes provider chat using the e2ee transport:
image

Chutes provider chat with tool calling and e2ee:
image

Chutes unit tests:
▶ Chutes Provider Config
  ✔ should exist in providerConfigs (0.326933ms)
  ✔ should expose TEE fallback models (0.401692ms)
✔ Chutes Provider Config (1.225487ms)
▶ Chutes E2EE Live Integration
  ✔ should be importable (10.624999ms)
  ✔ should discover available models (290.015261ms)
  ✔ should resolve model name to chute_id (65.794784ms)
  ✔ should fetch E2EE-capable instances (308.841838ms)
  ✔ should build encrypted request blob without plaintext leakage (276.006885ms)
  ✔ should complete non-streaming chat and return coherent English (8190.954085ms)
  ✔ should stream chat with valid SSE data lines (16927.810955ms)
  ✔ should work with reasoning model (Kimi-K2.5-TEE) (11961.162704ms)
✔ Chutes E2EE Live Integration (38031.624662ms)


agnt-gg:
Hey @alex-drocks, thank you so much for this contribution!! i love the chutes implementation and im pumped about the e2ee models! also i saw there was pliny GODMODE built in?? WHAT!? excited to use this provider

Thanks,
Nathan

Merged into main at 15bfc91, with your commit preserved as 4f9370d.

  Pre-merge fixes added on top:

  - ChutesDiscoveryManager: per-chute in-flight nonce-refresh dedup. Two concurrent first-use callers could each call _fetchInstances() and consume the first nonce from independent cache copies — protocol-level nonce-reuse risk on encrypted transport. Fix uses
  a Map<chuteId, Promise> so only one fetch runs and all waiters share the result.
  - modelTransform: emit undefined (not false) for missing supportsTools / reasoning / supportsVision arrays. The old behavior silently disabled tools on dynamic models that actually support them, since the downstream guard fires on === false.
  - chachaDecrypt: throw an explicit Chutes E2EE auth-failure error on null plaintext instead of letting it surface as a confusing gzip / Buffer coercion error several frames downstream.
  - appAuth.js: removed the hardcoded chutes entry in the local-fallback list now that it's a proper row in the remote auth_providers table — keeps the flow symmetric with every other apikey provider.
  - Reasoning toggle UX (Kimi-TEE / GLM-TEE / Qwen3-TEE) using chat_template_kwargs with per-family inner key (Kimi uses `thinking`, GLM and Qwen3 use `enable_thinking`).

  Tests added: concurrent same-chute distinct-nonce, parallel-chute isolation, drained-cache bounded retry, failed-refresh recovery, capability-array missing/empty/populated cases, tampered-ciphertext / tag / stream-chunk auth-failure assertions, plus
  per-family reasoning control + body-param tests.

  Closing this PR — the merge commit references it. Thanks again!

alex-drocks:
Hey @agnt-gg wow amazing that you found those improvements and fixed them.  Thank you so much.

PR #32, opened 2026-05-08
Fix desktop data path and mount tool schema routes

namle0251-blip:
## Summary

This fixes two desktop/runtime issues found while testing the Windows desktop build on AGNT 0.5.14:

- Prefer Electron `USER_DATA_PATH` before Docker `/app/data` when selecting the SQLite data directory.
- Mount the existing `ToolSchemaRoutes` router at `/api/tool-schemas` so documented tool-schema endpoints work.

## Why

In the packaged Windows desktop build, `/app/data` can resolve to `C:\app\data`. Because the current code checks that path before Electron `USER_DATA_PATH`, a desktop install can accidentally use a stale Docker-style data directory instead of `%APPDATA%\AGNT\Data`. This can make an update appear to wipe chat history by opening the wrong database.

The repo already contains `ToolSchemaRoutes.js` and docs for `/api/tool-schemas/*`, but the server does not mount the route. As a result `/api/tool-schemas/stats` returns 404.

## Changes

- Move the `USER_DATA_PATH` branch before Docker `/app/data`.
- Keep `/app/data` as a Docker/self-hosted fallback when `USER_DATA_PATH` is absent.
- Import and mount `ToolSchemaRoutes` in `backend/server.js`.

## Verification

Ran syntax checks locally:

```bash
node --check backend/server.js
node --check backend/src/models/database/index.js
```

Also verified the equivalent patched desktop build locally:

- `/api/health` returned `OK`
- `/api/version` returned `0.5.14`
- `/api/tool-schemas/stats` returned schema stats instead of 404
- live DB stayed under `%APPDATA%\AGNT\Data\agnt.db`
- `C:\app\data` was not recreated

### DIFF
--- a/backend/server.js
+++ b/backend/server.js
@@ -26,6 +26,7 @@ import GoalRoutes from './src/routes/GoalRoutes.js';
 import LayoutRoutes from './src/routes/LayoutRoutes.js';
 import OrchestratorRoutes from './src/routes/OrchestratorRoutes.js';
 import ToolsRoutes from './src/routes/ToolsRoutes.js';
+import ToolSchemaRoutes from './src/routes/ToolSchemaRoutes.js';
 import ModelRoutes from './src/routes/ModelRoutes.js';
 import CustomProviderRoutes from './src/routes/CustomProviderRoutes.js';
 import MCPRoutes from './src/routes/MCPRoutes.js';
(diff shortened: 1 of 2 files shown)


agnt-gg:
Hey @namle0251-blip — splitting your two findings apart:                                                                                                                                                                                                                             

**1. Missing `ToolSchemaRoutes` mount.** Real bug — file existed, docs referenced `/api/tool-schemas/*`, server.js never registered it. Landed in b90618f. You're now credited on the contributors graph via d39912a. Thanks for the catch.                                                                                                   

**2. Data-path precedence on Windows desktop.** You spotted a real symptom, but the underlying `database/index.js` cascade has since been replaced by `backend/src/utils/PathManager.js` (09693f2 / 3f16914), which gates the Docker tier on `process.platform !== 'win32'` —  Windows can't route to `/app/data` regardless of `USER_DATA_PATH` order. Pulling the reorder from this PR into the current tree would create a second cascade running parallel to PathManager, which would drift Docker users' paths between the two resolvers — quiet data-loss-shaped bug I want to avoid.

Both shipped in v0.5.15. Closing — the routes catch is a clean win on its own. 🙏

PR #33, opened 2026-05-08
Prefer local API keys for desktop auth

namle0251-blip:
## Summary

This improves desktop/local auth behavior and removes a token leak from Twitter/X health checks.

## Why

Some desktop integrations, especially local API-key integrations such as Obsidian Local REST API, store encrypted credentials in the local `api_keys` table. The current desktop auth flow often goes directly to AGNT hosted auth for `valid-token`/API-key retrieval, so local-only providers can appear disconnected even when the key is stored locally and valid.

The Twitter/X health check also logs the token value before calling `https://api.twitter.com/2/users/me`, which can leak credentials into app logs.

## Changes

- Make `AuthManager.getValidAccessToken` prefer local encrypted `api_keys` before falling back to hosted auth.
- Make the `agnt_auth` tool check local `api_keys` for `retrieve_api_key` and `get_valid_token`.
- Add Obsidian provider alias handling between `obsidian` and `obsidian-local-rest-api`.
- Include decoded user id when the AGNT SDK calls hosted `/auth/valid-token`.
- Stop logging Twitter/X tokens.
- Preserve useful Twitter/X health diagnostics and surface `client-forbidden` as an actionable API access/enrollment error.

## Verification

Ran syntax checks locally:

```bash
node --check backend/src/services/auth/AuthManager.js
node --check backend/src/libs/agnt2.js
node --check backend/src/services/orchestrator/tools.js
```

Also verified the equivalent patched desktop build locally with Obsidian Local REST API: the stored local API key was returned from `local_api_keys` and authenticated successfully against Obsidian's protected `/vault/` endpoint.

### DIFF
--- a/backend/src/libs/agnt2.js
+++ b/backend/src/libs/agnt2.js
@@ -600,10 +600,19 @@ class AuthModule {
     const response = await this.client.get(`/auth/connected`);
     return response.data;
   }
+  _getUserIdFromToken() {
+    try {
+      const payload = JSON.parse(Buffer.from(this.apiKey.split('.')[1], 'base64url').toString('utf8'));
+      return payload.id || payload.userId || payload.sub || null;
+    } catch {
+      return null;
+    }
+  }
   async getValidToken(providerId) {
-    // Assuming backend's /auth/valid-token uses req.user.id from the bearer token
-    // and providerId from query param.
-    const response = await this.client.get(`/auth/valid-token?providerId=${providerId}`);
+    const params = new URLSearchParams({ providerId });
+    const userId = this._getUserIdFromToken();
+    if (userId) params.set('userId', userId);
+    const response = await this.client.get(`/auth/valid-token?${params.toString()}`);
     return response.data; // Expected: { access_token: "..." }
   }
 }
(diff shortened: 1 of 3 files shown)


agnt-gg:
Hey @namle0251-blip — two of the things you flagged are legit:

 **1. Token leakage in `checkTwitterHealth` logs.** Right call — was logging the raw bearer. Fixed in e0ef61e (logs `token?.length` now).

 **2. "Twitter token validation failed" being unhelpful.** Also a fair point — message swallowed real signal. Root cause turned out to be Twitter's 250/24h cap on `GET /2/users/me` getting burned by routine health-check traffic, not missing app permissions. The fix in e0ef61e caches for 5 min, honors `x-user-limit-24hour-reset`, and only treats 401/403 as real validation failures.

You're now credited on the contributors graph via d39912a, covering both this PR and #32.

Passing on the broader local-first auth rewrite — the hosted-auth routing through `api.agnt.gg` is intentional so users don't need to register their own dev apps for every provider, and `.env` already covers fully-local creds for self-hosters. The Obsidian alias is better fixed by canonicalizing the id in the marketplace manifest. Also check out the new native Obsidian plugin release in 0.5.15!

Both shipped in v0.5.15. Closing — token-leak catch alone was a good one. 🙏

Issue #1, opened 2025-04-04
401 Incorrect api key provided. Unable to connect with OpenAI


wunluv:
I've set the .env file with my openai key. I;ve tested  key from terminal with openai_cli. Works fine. However the agent-js web app doesn't seem able to use the very same key that I'm using in the openai_sli
     
![Image](https://github.com/user-attachments/assets/5506a827-2334-4d43-8a9e-86a809244e1f)

wunluv:
Closing this. Deleting the key and adding a new key then restarting the server solved the issue

PR #34, opened 2026-05-08
[codex] Fix Codex context reduction for replay history

namle0251-blip:
## Summary

This PR fixes a Codex Responses context-management gap that can surface as `Your input exceeds the context window of this model` during long research/tool-heavy conversations.

## Root Cause

`CodexResponsesAdapter` was building requests from the full message history without running the existing `manageContext` reducer. Codex Responses also replays `_responsesOutputItems` for encrypted reasoning/tool state, but those replay payloads were not included in the token estimate, so AGNT could undercount large histories before sending them upstream.

## Changes

- Count `_responsesOutputItems` in `estimateMessagesTokens` so Codex replay state contributes to context budgeting.
- Run Codex Responses calls through preflight context management before building request params.
- Detect Codex context-window/token-limit errors and retry once with reduced context when reduction is possible.
- Add regression coverage for oversized Codex replay history and replay-payload token estimation.

## Validation

- `npm exec vitest run backend/src/services/orchestrator/llmAdapters.codex.test.js backend/src/utils/contextManager.test.js`
- `git diff --check`

### DIFF
--- a/backend/src/services/orchestrator/llmAdapters.codex.test.js
+++ b/backend/src/services/orchestrator/llmAdapters.codex.test.js
@@ -122,6 +122,51 @@ describe('CodexResponsesAdapter', () => {
     expect(result.responseMessage.content).not.toContain('Please check your OAuth connection');
     expect(result.responseMessage.content).not.toContain('reconnect your OAuth account');
   });
+
+  it('compresses oversized Codex replay history before streaming requests', async () => {
+    let capturedParams;
+    const client = {
+      responses: {
+        create: async (params) => {
+          capturedParams = params;
+          return streamFrom([
+            { type: 'response.output_text.delta', delta: 'OK' },
+            { type: 'response.completed', response: { id: 'resp_context', output: [], usage: {} } },
+          ]);
+        },
+      },
+    };
+    const adapter = await createLlmAdapter('openai-codex', client, 'gpt-5.5');
+    const hugeReplay = 'x'.repeat(1_300_000);
+
+    await adapter.callStream(
+      [
+        { role: 'system', content: 'System prompt' },
+        { role: 'user', content: 'Earlier research step' },
+        {
+          role: 'assistant',
+          content: '',
+          _responsesOutputItems: [
+            {
+              id: 'rs_huge',
+              type: 'reasoning',
+              summary: [],
+              encrypted_content: hugeReplay,
+              status: 'completed',
+            },
+          ],
+        },
+        { role: 'user', content: 'Continue with a concise answer' },
+      ],
+      [],
+      () => {},
+    );
+
+    const serializedParams = JSON.stringify(capturedParams);
+    expect(serializedParams).not.toContain(hugeReplay);
+    expect(serializedParams.length).toBeLessThan(100_000);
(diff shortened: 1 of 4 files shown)


agnt-gg:
Hey @namle0251-blip — drive-by feedback while this is still draft, hopefully saves you some time:

**Diagnosis is solid on three real gaps:**
1. `estimateMessagesTokens` in `contextManager.js` doesn't count `_responsesOutputItems`, so Codex's `encrypted_content` payload is invisible to the budget calculator — your one-line fix there is exactly right.
2. `CodexResponsesAdapter.call` / `callStream` don't have the token-limit recovery loop that every other adapter has (`OpenAiLikeAdapter` line 456, `AnthropicAdapter` line 1028, `CerebrasAdapter` line 2297, `GeminiAdapter` line 2570). Adding it brings Codex in line with the   rest.
3. Both compound — `manageContext` would undercount even if it did run, until #1 lands.

**One concern with the compression strategy** — the test that takes 1.3MB of `encrypted_content` down to <100KB by replacing it with a "Previous conversation" summary is the part I'd push back on. `encrypted_content` isn't human-readable history we can summarize; it's opaque ciphertext representing Codex's prior reasoning state, and the protocol expects it verbatim. Truncating or summarizing inside the blob is a Codex protocol violation waiting to happen — better to drop whole oldest turns (which `manageContext` already does for normal messages) than to touch encrypted_content internally.

 **Suggested split if you want to keep iterating:**
  - Keep #1 and #2 (token counting + recovery loop following the existing adapter pattern).
  - Drop the encrypted-content summarization — let `manageContext` handle reduction by removing oldest turns wholesale.

Happy to take the safe parts (#1 + #2) as a separate commit attributed to you so this doesn't block, while you decide if you want to rework or close. Either way, good catch on the gaps.

Annie @ AGNT

agnt-gg:
Thanks for the careful work here, @namle0251-blip. Your three-part diagnosis was exactly right and made the underlying fix much easier to land! Shipped in f9e8a08:                                                                   
                                         
**From your PR, directly:**                                                                                                                                                                                                             - `_responsesOutputItems` now counted in `estimateMessagesTokens` (the "exactly right" portion of your fix)
- Codex adapter retries on `context_length_exceeded` rejections, mirroring the pattern from your PR                                                                                                                                                                                               
**Reworked:**
- Instead of summarizing inside the `encrypted_content` blob, the retry sheds whole oldest atomic turns. The assistant message and its paired tool results drop together, which atomically removes the turn's `_responsesOutputItems`
  (including the encrypted blob) without ever reading or modifying its internals — same recovery effect, no protocol concerns.

**Added beyond your PR:**
- `compactMessageHistory` pass before every `manageContext` call, so historical bloat that bypassed per-tool offloading gets caught retroactively
- Removed the artifact-chat offload skip (was the proximate trigger of the symptom — base64-embedded HTML files)
- Fixed a separate cache-gate bug where successful reductions were being reverted because the gate compared post-management size instead of original

You're credited via `Co-Authored-By` on the commit. Closing this out since the issues are addressed — but please keep contributing, this kind of careful diagnosis is exactly what we want.

- Annie @ AGNT

PR #36, opened 2026-06-09
feat: add Crabbox validation support

zozo123:
## Summary
- Add a self-contained Crabbox plugin (`backend/plugins/dev/crabbox-plugin/`) exposing one workflow action for `run`, `warmup`, `status`, `inspect`, `list`, `stop`, `doctor`, `version`, `init`, `sync-plan`, and repo-local `job` flows, with parsed provider/lease/timing proof metadata.
- Zero changes outside the plugin folder — no committed build artifacts, no core docs/scripts edits. Distribution target is the plugin marketplace (`node build-plugin.js ./dev/crabbox-plugin`).

## Security model
- The Crabbox binary resolves from `CRABBOX_BIN` env / `crabbox` on PATH only — it is **not** a node parameter, so workflow authors cannot point the node at arbitrary host programs.
- The CLI child process gets a minimal env allowlist (`PATH`, `HOME`, `USERPROFILE`, temp dirs, `CRABBOX_*`) instead of inheriting the server environment; a unit test asserts server secrets never reach the CLI.
- Credentials route through AuthManager (`authRequired: "apiKey"`, `authProvider: "crabbox"`): the injected per-user token is passed as `CRABBOX_TOKEN`, and the node refuses to run on ambient `crabbox login` state.
- Remote commands are built argv-style (no local shell); `extraArgs` is argv-parsed and may not contain `--`.

## Lifecycle
- Default local timeout is 15 minutes; SIGTERM escalates to SIGKILL after a 10s grace period.
- If the local CLI is timed out or killed, the plugin parses the lease id and runs `crabbox stop <leaseId>` so the remote lease stops billing (skipped when `keep`/`keepOnFailure` is set); the attempt is reported in the `leaseCleanup` output.

## OpenClaw Crabbox alignment
- Verified against [crabbox.sh](https://crabbox.sh/) and `openclaw/crabbox` docs: `crabbox.sh` is the docs site; AGNT calls the local `crabbox` CLI.
- Supports `crabbox init --detect` style onboarding outputs (`.crabbox.yaml`, `.github/workflows/crabbox.yml`, `.agents/skills/crabbox/SKILL.md`) without vendoring repo-specific generated files.
- Supports repo-local jobs through `crabbox job list` and `crabbox job run <name>` for repeatable validation flows.

## House style
- `execute(params, inputData, workflowEngine)` contract; auth via `params.__auth.token` (same pattern as the ElevenLabs plugin).
- Boolean toggles are `inputType: "select"` with `["Yes","No"]`; ttl/idle defaults have one source of truth (90m/30m).
- Plugin-local `README.md` and tests under `backend/plugins/dev/crabbox-plugin/tests/`.

## Test plan
- `npx vitest run backend/plugins/dev/crabbox-plugin/tests` (17 tests: argv parsing, arg builder, Yes/No toggle handling, child-env allowlist, output/proof parsing)
- `node backend/plugins/build-plugin.js crabbox-plugin` (artifact generated locally, not committed)
- Earlier end-to-end smoke: `crabbox run --provider local-container ... -- "npm install --ignore-scripts && npx vitest run ..."` passed with provider `local-container`, lease `cbx_4cb5df261609`, exit `0`, `leaseStopped=true`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


backend/plugins/dev/crabbox-plugin/crabbox-api.js
```diff
@@ -0,0 +1,505 @@
...
+class CrabboxPlugin {
+  constructor() {
+    this.name = 'crabbox';
+  }
+
+  async execute(params = {}) {
+    console.log('[CrabboxPlugin] Executing with params:', JSON.stringify({ ...params, command: params.command }, null, 2));
+
+    try {
+      const binary = compact(params.binary) || process.env.CRABBOX_BIN || 'crabbox';
```
agnt-gg: 🔴 **Blocker — arbitrary host execution.** `binary` arrives from workflow params, so any workflow author can execute any program on the host. Drop the param entirely and resolve from server config only:

```js
const binary = process.env.CRABBOX_BIN || 'crabbox';
```

backend/plugins/dev/crabbox-plugin/crabbox-api.js
```diff
@@ -0,0 +1,505 @@
...
+
+  runProcess(binary, args, { cwd, timeoutMs, maxOutputBytes }) {
+    return new Promise((resolve, reject) => {
+      let stdout = '';
+      let stderr = '';
+      let timedOut = false;
+
+      const child = spawn(binary, args, {
+        cwd,
+        env: process.env,
```
agnt-gg: 🔴 **Blocker — secret leakage.** Your own CRABBOX.md says *“do not rely on inherited local environment as proof of intent”* — but this line hands the third-party CLI the server’s full env, including API keys and session secrets. Allowlist instead:

```js
const childEnv = {
  PATH: process.env.PATH,
  HOME: process.env.HOME,
  USERPROFILE: process.env.USERPROFILE,
  ...Object.fromEntries(Object.entries(process.env).filter(([k]) => k.startsWith('CRABBOX_')))
};
```

backend/plugins/dev/crabbox-plugin/crabbox-api.js
```diff
@@ -0,0 +1,505 @@
...
+        shell: false,
+      });
+
+      const timer = setTimeout(() => {
+        timedOut = true;
+        child.kill('SIGTERM');
```
agnt-gg, line 682: 🟡 **Cost hazard — remote lease leaks on timeout.** This kills the local CLI, but the cloud lease keeps running (and billing) on the broker side. Track the lease ID from stdout and run `crabbox stop <leaseId>` in a `finally`, escalate SIGTERM→SIGKILL after a grace period, and reconsider the 2-hour default timeout for a workflow node (10–15 min is saner).

.dockerignore
```diff
@@ -97,6 +97,7 @@ build/
 
 # Plugin builds (will be generated in container)
 backend/plugins/plugin-builds/*.agnt
+!backend/plugins/plugin-builds/crabbox-plugin.agnt
```
agnt-gg: ⛔ **Must revert.** The comment two lines up says *“will be generated in container”* — that’s the design. This exception exists only to carry a committed binary `.agnt` into Docker images, and `plugin-builds/` is gitignored precisely so builds are never committed. Both this line and `plugin-builds/crabbox-plugin.agnt` need to go.

agnt-gg (changes requested):
## Thanks for this! 🦀

First — genuinely appreciate the effort here. The arg-builder/parser code is clean, you shipped real unit tests, and you clearly read the plugin docs (conditional params, output shapes, ENOENT handling — all correct). We’d like to see Crabbox available to AGNT users — but this PR can’t merge in its current shape. Here’s exactly why, and exactly what the path forward is, because there absolutely is one.

### The core issue: plugins must be 100% self-contained

AGNT’s plugin system is designed so an integration lives entirely in **one folder**. This PR touches 6 paths outside the plugin folder, and each one is a problem:

| File | Issue |
|---|---|
| `plugin-builds/crabbox-plugin.agnt` | ⛔ Committed build artifact. This dir is gitignored for a reason — builds are generated, never committed. An unreviewable binary that would ship inside every desktop build (via `asarUnpack`) is a non-starter. |
| `.dockerignore` | ⛔ Exists only to carry the artifact above into Docker images. Must revert. |
| `docs/CRABBOX.md` | Move into `dev/crabbox-plugin/README.md` (Best Practice #4 in `PLUGIN-DEVELOPMENT.md`). |
| `README.md` table row | Revert — the main docs table is for core platform docs. |
| `scripts/crabbox.sh` | Remove — repo `scripts/` is build tooling only, and a bash-only wrapper doesn’t fit a Windows-first app. |
| `tests/unit/crabbox-*.test.js` | Move alongside the plugin. |

**Target structure:**
```
backend/plugins/dev/crabbox-plugin/
├── manifest.json
├── crabbox-api.js
├── package.json
├── README.md      ← your docs/CRABBOX.md content goes here
└── tests/
```

### Security blockers (must fix regardless of where this lands)

1. **`binary` cannot be a node parameter.** Workflow params are user/agent-controlled — `spawn(params.binary, ...)` makes this an “execute any program on the host” node. Resolve the binary from `CRABBOX_BIN` env/server config only. (Inline comment on the exact line.)
2. **Don’t pass `env: process.env`** to the child. The server env contains API keys and secrets. Build a minimal allowlist (`PATH`, `HOME`, `CRABBOX_*`).
3. **Use the documented tool contract**: `async execute(params, inputData, workflowEngine)`, and route Crabbox credentials through AuthManager (`authRequired`/`authProvider`) instead of ambient CLI login — otherwise every user on a shared instance executes as the server. Worked example in `PLUGIN-DEVELOPMENT.md` § Authentication.
4. **Lease cleanup**: on timeout you `SIGTERM` the local CLI but the remote lease keeps billing. Call `crabbox stop <leaseId>` in the failure path, escalate SIGTERM→SIGKILL, and drop the 2-hour default timeout to something sane (10–15 min).

### Distribution: this should be a marketplace plugin

Honest expectations: third-party service integrations ship via the **plugin marketplace**, not as vendored defaults in core (that’s why nothing in this PR should touch core files). The flow is: keep the source self-contained as above → `node build-plugin.js ./dev/crabbox-plugin` → publish the `.agnt` via the marketplace. We’re happy to help with that process — it gets Crabbox in front of every AGNT user *without* needing core-repo changes for each release.

### Checklist to approval

- [ ] All files inside `backend/plugins/dev/crabbox-plugin/` — zero changes to `.dockerignore`, `README.md`, `docs/`, `scripts/`, `plugin-builds/`
- [ ] No committed `.agnt` artifact
- [ ] `binary` removed from node params (env/config only)
- [ ] Minimal child env (no `process.env` passthrough)
- [ ] `execute(params, inputData, workflowEngine)` signature + AuthManager for credentials
- [ ] Remote lease cleanup on timeout/error; saner default timeout
- [ ] Plugin-local README; `author` field reflects the actual author (that’s you — take credit!)
- [ ] Bonus: `inputType: "select"` + `["Yes","No"]` for boolean toggles (house style), and align the ttl/idle defaults (manifest says 90m/30m, the shell wrapper said 120m/60m — one source of truth)

Happy to answer questions on any of these. Looking forward to v2 🚀

zozo123:
Thanks for the detailed review — all of it made sense, and everything on the checklist is addressed in 93236ca. The PR now touches exactly 5 files, all inside `backend/plugins/dev/crabbox-plugin/`.

- [x] **Self-contained**: reverted `.dockerignore` and the `README.md` table row; removed `scripts/crabbox.sh` and `docs/CRABBOX.md` (content now lives in the plugin's `README.md`); tests moved to `backend/plugins/dev/crabbox-plugin/tests/`
- [x] **No committed `.agnt` artifact**: `plugin-builds/crabbox-plugin.agnt` is gone; build instructions are in the plugin README
- [x] **`binary` removed from node params**: resolved from `CRABBOX_BIN` env / `crabbox` on PATH only
- [x] **Minimal child env**: allowlist of `PATH`, `HOME`, `USERPROFILE`, temp dirs, and `CRABBOX_*` — no `process.env` passthrough (covered by a unit test that asserts server secrets never reach the CLI)
- [x] **Documented contract + AuthManager**: `execute(params, inputData, workflowEngine)`, schema declares `authRequired: "apiKey"` / `authProvider: "crabbox"`, and the injected `params.__auth.token` is passed to the CLI as `CRABBOX_TOKEN` per user. The node now refuses to run on ambient CLI login — happy to relax that to a warning if you'd rather keep direct-provider/`local-container` usable without a broker credential.
- [x] **Lease cleanup**: on timeout/kill the plugin parses the lease id and runs `crabbox stop <leaseId>` (skipped when `keep`/`keepOnFailure` is set), reports the attempt in a `leaseCleanup` output field, escalates SIGTERM→SIGKILL after a 10s grace, and the default local timeout is now 15 min
- [x] **Plugin-local README + author field** updated
- [x] **Bonus items**: all boolean toggles are now `inputType: "select"` with `["Yes","No"]`, and ttl/idle defaults have one source of truth (90m/30m — the conflicting shell wrapper is deleted)

On distribution: agreed marketplace is the right home — once the source shape looks good to you I'll build the `.agnt` and follow the marketplace publishing flow. Would appreciate pointers on hosting/`marketplace.json` when we get there. 🦀

🤖 Generated with [Claude Code](https://claude.com/claude-code)

backend/plugins/dev/crabbox-plugin/crabbox-api.js
zozo123: Fixed in the latest push: failed or timed-out RUN/WARMUP paths now parse the lease ID and run `crabbox stop <leaseId>` unless the lease was caller-supplied or explicitly kept; local timeout defaults to 15m with clamping, SIGTERM escalates to SIGKILL, and plugin tests cover the cleanup decisions.

backend/plugins/dev/crabbox-plugin/crabbox-api.js
zozo123: Fixed. The `binary` param is gone entirely — `resolveBinary()` now resolves only from `CRABBOX_BIN` (validated to be an existing file) or a PATH lookup, always to an **absolute** path before `spawn`. On Windows it additionally refuses `.cmd`/`.bat` shims (which would need a shell) and never resolves a bare name against the workflow-controlled cwd. No workflow param can influence which executable runs.

backend/plugins/dev/crabbox-plugin/crabbox-api.js
zozo123: Fixed. `process.env` is never passed to the child. `buildChildEnv()` builds a minimal allowlist — `PATH`, `HOME`, `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`, `SystemRoot`, temp dirs, plus explicit `CRABBOX_*` settings — and nothing else. `CRABBOX_TOKEN` is injected solely from the per-run AuthManager credential and is explicitly excluded from the ambient-env copy, so server API keys/secrets never cross the boundary.

.dockerignore
zozo123: Reverted. Both the `.dockerignore` exception and the committed `plugin-builds/crabbox-plugin.agnt` artifact are gone. The PR now touches **only** the five files inside `backend/plugins/dev/crabbox-plugin/` — `git diff --name-only upstream/main...HEAD` confirms zero changes to `.dockerignore`, `README.md`, `docs/`, `scripts/`, or `plugin-builds/`.

zozo123:
## v2 — every checklist item addressed 🦀

Thanks again for the incredibly detailed review. Everything is now resolved and the PR is ready for another look. Mapping to your checklist:

- [x] **All files inside `backend/plugins/dev/crabbox-plugin/`** — the PR now changes exactly 5 files, all in the plugin folder. `git diff --name-only upstream/main...HEAD` shows **zero** changes to `.dockerignore`, `README.md`, `docs/`, `scripts/`, or `plugin-builds/`.
- [x] **No committed `.agnt` artifact** — removed, along with the `.dockerignore` exception that carried it.
- [x] **`binary` removed from node params** — resolved from `CRABBOX_BIN`/PATH only, always to an absolute path, with Windows `.cmd`/`.bat` shim rejection. No param can influence the executable.
- [x] **Minimal child env** — `buildChildEnv()` allowlists `PATH`/`HOME`/temp/system dirs + explicit `CRABBOX_*`; `process.env` is never passed through. `CRABBOX_TOKEN` comes only from the per-run credential.
- [x] **`execute(params, inputData, workflowEngine)` + AuthManager** — manifest declares `authRequired: "apiKey"` / `authProvider: "crabbox"`; the token arrives via `__auth` and the node hard-errors if no credential is connected. Ambient CLI login is never used.
- [x] **Remote lease cleanup + saner timeout** — `cleanupLease()`/`shouldStopLease()` run `crabbox stop <leaseId>` in the failure/timeout path (skipping caller-supplied or explicitly-kept leases), SIGTERM escalates to SIGKILL after a grace period, and the default timeout dropped from 2h to **15m** (60m cap).
- [x] **Plugin-local README + `author`** — `README.md` lives in the plugin folder; `author` is `Yossi Eliaz`.
- [x] **Bonus** — all boolean toggles are `inputType: "select"` with `["Yes","No"]`; ttl/idle now have a single source of truth in the manifest (`90m`/`30m`) since the shell wrapper is gone.

All **27 unit tests pass** (`vitest run`), and `mergeable` is clean against `main`.

Re-requesting review — happy to follow up on marketplace publishing once this lands. 🚀

agnt-gg:
@zozo123 — thank you again for the careful follow-through here. The latest version closes the review blockers, keeps the plugin self-contained, removes the committed artifact, adds the safer process/env handling, and the test coverage looks solid. The extra hardening you added around binary resolution, env isolation, and lease cleanup is exactly the kind of security instinct we want in the plugin ecosystem.

I also owe you an apology for the confusion around **where plugins should live** and how marketplace publishing works.

That confusion is on our docs and review process, not on you.

The old plugin docs blurred two different paths:

1. **AGNT-bundled/default plugins** — plugins we ship or list from the core repo.
2. **Third-party/user-published plugins** — plugins you build, publish, and own through the marketplace.

Your PR followed the docs as they existed, but for a third-party integration like Crabbox, the better path is the second one: **publish it as a marketplace plugin under your own account**, rather than merging it into the core repo.

We’ve updated the docs to clarify the distinction:

- `marketplace.json` is only for AGNT-bundled/default marketplace entries.
- Third-party/user-published plugins do **not** edit `marketplace.json`.
- You do **not** need a core PR to publish a plugin.
- The normal flow is: build the `.agnt`, then publish it through the marketplace UI.

For Crabbox, from your branch, build the package with:

```bash
node build-plugin.js ./backend/plugins/dev/crabbox-plugin
```

That produces:

```txt
crabbox-plugin.agnt
```

Then publish it in the app through either path:

1. **Marketplace page → Plugins tab → right-side “Publish” panel**
2. **Connectors tab → Plugins screen → Publish tab**

Publishing it yourself is the better outcome here because the listing belongs to your account. You’ll be able to edit the listing, update metadata, and publish new versions without waiting on core repo review or AGNT release cadence.

On the credential behavior you mentioned — whether the node should hard-error without a broker credential, warn, or allow local/direct-provider use without one — that’s your call as the plugin owner. AuthManager gives users a place to store secrets for the plugin, but a plugin can also support env vars, local modes, or no auth depending on what makes sense for the integration. So don’t let that block publishing; choose the behavior that fits Crabbox best.

So my recommendation is:

- publish Crabbox as a marketplace plugin under your account,
- drop the marketplace link here once it’s live,
- and then we can close this PR as **shipped via marketplace**, not rejected.

To be clear: the work is good. The PR just ended up being the wrong distribution path because our docs didn’t explain the distinction clearly enough. Thanks for helping expose that gap, and thanks for the quality of the plugin work.

agnt-gg (changes requested):
This looks great!!! Check Annie's latest feedback on how to publish!!:
https://github.com/agnt-gg/agnt/pull/36#issuecomment-4737510672

Ping me in discord if you need help building the plugin to a .agnt file and publishing it to the marketplace! 

If you already have the plugin installed in your own AGNT instance its super easy, just go to the Marketplace > Plugins tab and "Publish" on the right panel: See Pic:

Image

Once i know for sure we've got your plugin published under your account we can close this PR as completed successfully!

- Nathan

zozo123:
Thanks again @agnt-gg for the thorough review and the clarification on where third-party plugins belong. Per your guidance, Crabbox is a third-party integration that ships through the **marketplace** rather than as a core-repo merge — so I'm closing this PR rather than leaving it open against `main`.

The work is complete and self-contained: all review blockers addressed, plugin lives entirely in `backend/plugins/dev/crabbox-plugin/`, no committed artifact, hardened binary resolution / env allowlist / lease cleanup, and the build packages cleanly (`node build-plugin.js crabbox-plugin` → `crabbox-plugin.agnt`). I'll publish it to the marketplace under my own account from there.

Appreciate the help and the friendly process — 🦀🚀

PR #38, opened 2026-06-17
feat/unfirehose docker mount & bash exec in docker

russellballestrini:
Screenshot from 2026-06-17 15-53-15

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -90,7 +90,11 @@ RUN apk add --no-cache \
     ttf-freefont \
     sqlite \
     ffmpeg \
-    su-exec
+    su-exec \
+    bash \
+    git \
+    curl \
+    jq
 
 # Set Puppeteer to use installed Chromium
 ENV PUPPETEER_SKIP_DOWNLOAD=true
--- a/backend/src/services/OrchestratorService.js
+++ b/backend/src/services/OrchestratorService.js
@@ -24,7 +24,7 @@ import UserModel from '../models/UserModel.js';
 import AgentModel from '../models/AgentModel.js';
 import SkillModel from '../models/SkillModel.js';
 import { buildSkillsContext } from './SkillService.js';
-import { createSession as createUnfirehoseSession, wrapSendEvent as wrapUnfirehoseSendEvent, isEnabled as isUnfirehoseEnabled } from './unfirehose/UnfirehoseLogger.js';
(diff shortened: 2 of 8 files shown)


agnt-gg:
Hey Russell — Annie here! 👋

This is looking great after the force-push. I like the tighter scope a lot: persisting unfirehose logs out of Docker, giving sessions better per-entity project slugs, and making the full Docker image friendlier for real shell workflows with `bash/git/curl/jq`.

The Docker mount + entrypoint changes match the existing AGNT pattern nicely — especially the `/app/unfirehose` self-heal alongside `/app/logs` and `/app/data`. The `deriveProjectSlug()` helper is also a nice shape: pure, easy to test, and wired in at the single unfirehose session creation point.

One small path-hardening nit before merge:

In `backend/src/services/unfirehose/UnfirehoseLogger.js`, most values that can become part of the unfirehose project folder go through `slugifyName()`, but the `goalId` fallback currently uses the raw string prefix:

```js
if (goalId) return `goal-${String(goalId).slice(0, 8)}`;
```

Since this value ultimately becomes part of the output directory, I’d run the fallback through the same slug helper for consistency:

```js
if (goalId) {
  const id = slugifyName(goalId);
  if (id) return `goal-${id.slice(0, 8)}`;
}
```

Would also be nice to add a tiny test for that fallback, e.g. a goal id with path-ish characters, just to lock in the behavior.

After that, I think this is good to go. Suggested final smoke check would be:

- slug tests pass
- Docker starts cleanly
- `/app/unfirehose` is writable
- host `~/.agnt/unfirehose` receives the JSONL session files

Excited for this one — persistent unfirehose logs from Docker are going to make those sessions way easier to inspect.

russellballestrini:
@agnt-gg hey annie, take another pass?

agnt-gg (approved):
Perfect — that's exactly it. 🙌

Traced the new tests against `slugifyName` by hand and they all check out:

- `../etc/passwd` → `goal-etc-pass` (the hostile path-traversal case I wanted)
- `foo/bar/baz` → `goal-foo-bar-`
- empty-after-slugify → falls through to `chat` — nice catch adding that guard, I didn't even call it out

Slugify-first-then-slice is the right shape, and nothing can reach `join(UNFIREHOSE_DIR, projectSlug)` with path-ish characters anymore. Test count up to 14.

Everything else (Docker mount, entrypoint self-heal, `bash/git/curl/jq`, OrchestratorService wiring) is unchanged from my last pass and already looked great.

Approving — thanks for being so quick and thorough on the turnaround, Russell. Excited to have persistent unfirehose logs from Docker. 🚀

PR #39, opened 2026-06-18
Fix/router auth redirect visibility

russellballestrini:


frontend/src/router/authGuard.js
```diff
@@ -0,0 +1,86 @@
...
+      next({
+        path: '/connectors',
+        query: to.query,
+      });
+      return;
+    }
+
+    if (to.meta.requiresAuth && !storeInstance.state.userAuth.user) {
+      try {
+        await storeInstance.dispatch('userAuth/fetchUserData');
```
agnt-gg: 🔴 **Blocker — the auth gate rides the freshness cache.**

`fetchUserData` is wrapped in `withFreshness(..., { staleAfter: TTL.userAuthFetchUserData })`, so when a probe ran recently this dispatch returns the cached no-op and the action body never executes — meaning the guard reads a stale `lastAuthFailure` and never re-probes.

Two consequences for the gate specifically:
- **Sticky transient failure:** after an `http_5xx`, a retry within the TTL keeps showing “service is down” even once the service is back.
- **Delayed token pickup:** a token that becomes valid mid-TTL isn’t honored until the cache expires.

An active attempt to enter a gated route should trigger a live auth check, which is exactly the documented escape hatch:

```js
await storeInstance.dispatch('userAuth/fetchUserData', { forceRefresh: true });
```

If you intentionally wanted the gate to ride the cache, let me know the reasoning and I’ll drop this. 🙂

agnt-gg (changes requested):
Hey Russell 👋 — read through the whole thing and honestly this is really nice work. The factory extraction so tests don’t drag in the app graph, the 14 specs, the transient-vs-definitive token call, even sweeping up that pre-existing `sounds-settings-changed` listener leak in `onUnmounted` — all the right instincts. The commit messages pinning the contract are *chef’s kiss*.

One real thing before I give it the green light — left it inline on the `fetchUserData` dispatch in `authGuard.js`.

### The blocker: the gate rides the freshness cache

`fetchUserData` is wrapped in `withFreshness(..., { staleAfter: TTL.userAuthFetchUserData })`, so if a probe ran recently this dispatch short-circuits to the cached no-op and the action body never runs. Perfect for a background refresh — but this is the **gate**, a user actively trying to walk through a protected door. Riding the cache here bites in two spots:

1. **Sticky transient failures** — a probe records `http_5xx`, the service recovers, but a retry inside the TTL window still shows “service is down” because the guard reads the stale `lastAuthFailure` instead of re-checking.
2. **Delayed token pickup** — a token that goes valid mid-TTL might not get honored on the next gated nav until the cache expires.

The wrapper docblock literally calls this out as the escape hatch, so I think it just wants `{ forceRefresh: true }`. Active attempt at a gated route = real auth probe, not a cached one. Details on the inline comment.

### Optional tidies (neither blocking)

- `TerminalLayout` pipes `detail.from` (= `to.fullPath`) into the modal, which `SimpleModal` renders via `v-html`. Practically safe since `fullPath` is encoded, but `to.path` would be cleaner and drops the `?content-id=…` noise from the user-facing copy.
- The `unauthenticated_response` branch records the failure but skips `SET_USER(null)` — harmless now, just keeps both paths symmetric.

That’s it. Flip the `forceRefresh` and I’ll re-review same day. And if you *meant* to let the gate ride the cache, tell me why and I’ll happily drop the gate instead. 🙂

— Annie

russellballestrini:
@agnt-gg 

Changes:
  1. authGuard.js — gate dispatches fetchUserData with { forceRefresh: true } to bypass the withFreshness TTL cache. Live probe, not background refresh. Header docblock + inline comment explain why.
  2. authGuard.js — event detail.from switched from to.fullPath → to.path (drops ?content-id=… noise from user-facing modal copy). returnTo still uses fullPath so deep-link resume preserves query. Warn log still uses fullPath so admins can grep the exact route.
  3. userAuth.js — unauthenticated_response branch now also commits SET_USER(null), symmetric with the happy path's SET_USER(user).
  4. authGuard.spec.js — pinned contract docblock updated, existing tests adjusted for new from shape, new test dispatches fetchUserData with forceRefresh:true so a future revert fails CI.

agnt-gg (approved):
Hey Russell 👋

v2 looks good — all three points addressed cleanly.

`forceRefresh` on the gate dispatch is correct and well-documented. The `to.path` / `to.fullPath` separation for user-facing vs admin-facing copy is a smart call — better than what I originally suggested. `SET_USER(null)` symmetry is clean.

Also noticed you fixed the `sounds-settings-changed` listener leak in TerminalLayout — that was a pre-existing bug nobody caught. Nice eye.

Tests pin everything that matters. Approving.

— Annie 🤖

PR #41, opened 2026-07-02
[codex] Add button press and release sounds

TheSeanLavery:
## Summary
- add separate button press and release sounds for terminal UI interactions
- preload UI audio assets before playback to reduce first-click latency
- keep explicit `data-sound` click behavior while avoiding the delayed default click fallback

https://github.com/user-attachments/assets/24edd25d-f80c-4cbd-b576-8674bff55a34

## Scope
This is intentionally limited to `TerminalLayout.vue` plus the two referenced WAV files. Local scratch files are not included.

## Validation
- `npm run build` from `frontend`
- `git diff --check`


agnt-gg:
Hey Sean 👋

Ran this on my end — a couple of blockers.

**Performance:** `main` currently has one `click` listener playing one sound (`mouse-click.mp3`). This PR adds two more listeners (mousedown + mouseup) each doing `new Audio()` per event. Every button tap fires three listeners and potentially three sounds. That's what's causing the hang — `new Audio(src)` is a DOM element allocation + audio pipeline setup, and it stacks up fast during normal interaction.

**The press sound (`in.wav`) is harsh.** Sharp attack transient — even at moderate volume it's jarring compared to the existing `mouse-click.mp3`. Needs a softer sample or significantly lower volume.

**The release sound (`out.wav`) adds noise without value.** Users don't consciously register a separate mouseup sound — it just makes every interaction feel twice as busy. One soft press sound is enough.

I'd hold this one until #42 lands (which fixes the audio engine and supersedes the changes here). Happy to review that one when it's ready.

— Annie 🤖

agnt-gg (commented):
Sean, this is such a nice detail — the two-stage press/release feedback genuinely makes the UI feel more alive, and I can tell you sweated the first-click latency. Love it.

A few perf spots I'd like to see tightened before it lands — partly because I think this component is close to what surfaced the whole perf thread in the first place, so getting it right here pays off twice:

1. **`preloadSounds()` at mount loads all 7 sound files up front** — `new Audio()` + `.load()` for every entry the moment TerminalLayout mounts. You already track first interaction with `setInteracted`, so what if we preload just `buttonDown`/`buttonUp` there and lazy-load the rest on first use? Keeps the snappiness, skips the boot burst.
2. **The mousedown/mouseup capture listeners fire on every mouse event** — drags, scrolls, selections all hit `getSoundTrigger()` and its DOM walk. An early-return on the sound-disabled flag *before* the queries makes it a clean no-op when sound's off.
3. **`cloneNode(true)` per click** still creates a media element each time. For short blips, Web Audio (`decodeAudioData` once → `createBufferSource` per play) is the sweet spot and dodges the pile-up entirely. Happy to pair on it — it's a fun little rewrite.

Really solid feature, just want to keep it weightless at boot. 🙏

**Rec: keeping this open — no conflicts with anything on main.** Address the notes (even just #1 and #2 if you want to keep it small) and I'll review + merge. This one doesn't need to wait on the others.

PR #42, opened 2026-07-02
[codex] Improve terminal screen loading performance

TheSeanLavery:
## Summary
- lazy-load the terminal route and OAuth callback route instead of importing them into the initial router bundle
- lazy-load Chat and Settings screens, then load only the active screen immediately
- replace preload-every-screen behavior with section-scoped idle preloading for likely next screens
- bound terminal screen `KeepAlive` cache to reduce long-session memory retention
- switch UI sound playback to decoded Web Audio buffers with HTML audio fallback

## Scope
This PR intentionally covers performance items 2-5 from the local review: screen preload policy, cache bounds, Chat/Settings laziness, and Web Audio playback. The global `index.html` library cleanup is left out so this PR stays focused.

## Validation
- `git diff --check`
- `npm ci` in `frontend`
- `npm run build` in `frontend`
- preview smoke on `http://127.0.0.1:5301/settings` with Playwright

## Smoke Result
The `/settings` preview loaded without page errors. The probe confirmed Settings no longer loaded the Chat chunk, vendor-3d, vendor-editor, or vendor-charts. Remaining preview console errors were expected because Vite preview does not serve the backend API/socket routes.

### DIFF
--- a/frontend/src/router/index.js
+++ b/frontend/src/router/index.js
@@ -1,9 +1,9 @@
 import { createRouter, createWebHistory } from 'vue-router';
 // import Marketplace from '@/views/Marketplace/Marketplace.vue';
 // import ExecutionDetails from '@/views/ExecutionDetails/ExecutionDetails.vue';
-import Terminal from '@/views/Terminal/Terminal.vue';
+const Terminal = () => import('@/views/Terminal/Terminal.vue');
 const DocsView = () => import('@/views/Docs/Docs.vue');
-import OAuthCallback from '@/views/_components/utility/OAuthCallback.vue';
+const OAuthCallback = () => import('@/views/_components/utility/OAuthCallback.vue');
 import store from '@/store/state';
 
 const router = createRouter({
--- a/frontend/src/views/Terminal/Terminal.vue
+++ b/frontend/src/views/Terminal/Terminal.vue
@@ -11,7 +11,7 @@
       @screen-change="changeScreen"
     >
       <!-- KeepAlive caches visited screens so charts/data don't reload on every navigation -->
-      <KeepAlive>
+      <KeepAlive :max="4">
         <component
           v-if="isScreenReady"
           :is="activeScreenComponent"
@@ -37,7 +37,7 @@
 </template>
 
 <script>
-import { ref, computed, onMounted, watch, defineAsyncComponent, shallowReactive, markRaw, nextTick } from 'vue';
+import { ref, computed, onMounted, watch, shallowReactive, markRaw } from 'vue';
 import { useRoute, useRouter } from 'vue-router';
 import { useStore } from 'vuex';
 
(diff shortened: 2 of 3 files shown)


agnt-gg:
Hey — thanks for looking into this! Couple of high-level thoughts before you go further:

- **Chat is the default screen** and nearly every session lands on `/` first, so lazy-loading it would add a request waterfall to the critical path. The heavy vendors (Three.js, Chart.js) are already lazy-imported at point-of-use inside `MessageItem.vue`, so they don't actually block the entry bundle today — what the smoke test picked up disappearing from `/settings` was the idle prefetch warming those chunks in the background, not a critical-path savings.

- **The Web Audio change is solid thinking** but probably better as its own PR — easier to review and test in isolation (there's also a subtle behavior change where sounds now overlap on rapid clicks instead of stopping the previous one, worth a deliberate call either way).

- **`KeepAlive :max="4"` and lazy `OAuthCallback`** are easy wins we'd take as a small standalone PR anytime.

Let me know if you have questions about the bundle architecture — happy to point you at the parts that matter most.

— Annie

agnt-gg (changes requested):
Sean, thank you for going deep on this — you clearly dug into the real loading behavior instead of guessing, and it shows. A couple of these are things I flat-out missed.

Heads up: I landed PRD-105 (startup perf, safe batch) on main earlier today in `65331dc`, and it overlaps you in a couple spots. Specifically I made Settings async in `Terminal.vue`, so that piece of yours is now redundant and will conflict there.

Two things in here I genuinely want to keep:
- **Router-level lazy Terminal + OAuthCallback** — honestly a great catch, that was a real gap on my side. Peel it into its own tiny PR and I'll fast-track it.
- **Bounded KeepAlive** — proper memory-hygiene win, worth its own PR too.

Two I'd like to park until we have numbers:
- **Lazy-loading Chat** — since it's the default screen, deferring it adds a waterfall before first paint. I'd love the #45 harness to weigh in before we commit.
- **The idle-preload rewrite of TerminalLayout** — the current preload-after-Chat is holding up fine, so let's make that call with data rather than swap it blind.

**Rec: rebase and split — not a close.** Router lazy-load into its own tiny PR (I'll merge that basically on sight), KeepAlive into a second, and drop the Settings change since it's already in. The lazy-Chat + preload-rewrite bits I'd shelve for now — not because they're wrong, but because #45 is about to give us the data to judge them properly. I'm marking Request Changes only so the merge button stays locked until the rebase — everything here is landing on the new main, nothing's rejected.

agnt-gg:
Hey Sean — I'm closing this one out, and I want to be clear about why, because parts of it were genuinely good.

**The core issue is scope.** This PR fuses four separate changes into one commit: the router lazy-load, the Chat screen laziness + preload-policy rewrite, the KeepAlive bound, and a Web Audio rewrite of `playSound`. We hold a hard line on this: **one concern per PR**, so each change can be reviewed, measured, and reverted independently. When a 3-file diff contains four behavioral changes, a regression in any one of them takes the whole thing down and nobody can bisect it.

Where each piece landed:

- **Router lazy `Terminal` + `OAuthCallback`** ✅ — this is the gem. Standard pattern, zero behavior change, real entry-chunk win. It's tracked internally and will land as its own commit.
- **Lazy Settings** — already shipped on main (`65331dc`, `defineAsyncComponent`), so this part now conflicts with main for a change that's done.
- **Lazy Chat + section-scoped idle preloading** ❌ — Chat is the default screen; deferring it puts a load waterfall *in front of first paint*. And the preload rewrite replaces working behavior with new machinery with no before/after numbers attached. We don't take speculative behavior changes without measurement.
- **`KeepAlive :max="4"`** — legit memory-hygiene idea, but eviction changes behavior (evicted screens re-mount and re-fetch), and `4` is an unexplained magic number. Worth doing, but as its own reviewed commit — tracked internally.
- **Web Audio playSound** — see #41; the sound feature is being decided on its own merits there, and an audio-engine rewrite doesn't belong inside a startup-perf PR either way.

Also worth noting: a lot of the baseline sluggishness that motivated this PR has since been fixed on main from other angles:

- `65331dc` — PRD-105 startup batch: startup JS graph 2,488 KB → 1,420 KB (−43% gzip), zero CDN requests at first paint, flat 250ms backend polling (−2–4s launch), single-pass tool registration
- `211bb84` — workflow canvas perf + **GPU acceleration re-enabled** (this one matters a lot for general UI/interaction smoothness)
- `4798cbe` — dashboard activity chart query was O(entire history) per load; now covering-index + lazy daily rollup
- `4bf7b50` — failed activity fetches back off instead of hot-retrying
- `fc40f22` — model lists now SWR-cached with persistent fallback instead of blocking fetches

Worth pulling latest main and re-feeling the app before investing more in this direction — the profile that motivated this PR has shifted quite a bit underneath it.

Nothing here was wasted — the good ideas are captured. Going forward: if you'd sent the router change as a standalone 2-line PR, it would have merged the same day. That's the shape we want. Small, single-purpose, measured.

— Annie

PR #44, opened 2026-07-02
[codex] Lazy-load global frontend libraries

TheSeanLavery:
## Summary
- Stacked on #42. This PR should be reviewed after #42, or rebased once #42 lands.
- Removes eager startup script tags for highlight.js, MathJax, Showdown, html2canvas, jsPDF, and jsPDF autotable.
- Adds a shared lazy global-library loader and updates markdown, math, highlighting, and PDF export call sites to load only when needed.

## Validation
- git diff --check
- npm run build
- Vite preview smoke on http://127.0.0.1:5301/settings confirmed the app shell loads and the heavyweight globals are absent on initial load.

## Notes
- Preview-only backend/auth/socket console errors are expected when serving the built frontend without the AGNT backend.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -6,38 +6,6 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>AGNT.gg</title>
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.1/css/all.css" />
-    <script id="Highlight-script" src="/js/libs/highlight.js"></script>
-    <script defer>
-      window.MathJax = {
-        tex: {
-          inlineMath: [['\\(', '\\)']],
-          displayMath: [
-            ['\\[', '\\]'],
-            ['$$', '$$'],
-            // ['$', '$'],
-          ],
-          processEscapes: true,
-          processEnvironments: true,
-          packages: { '[+]': ['mhchem'] },
-        },
-        svg: {
-          fontCache: 'global',
-        },
-        loader: {
-          load: ['[tex]/mhchem'],
-        },
-        startup: {
-          typeset: false,
-          pageReady: () => {
-            return MathJax.startup.defaultPageReady();
-          },
-        },
-      };
-      hljs.configure({
-        ignoreUnescapedHTML: true,
-      });
-    </script>
-    <script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js" defer></script>
     <script defer>
       // Add this script at the end of the head section
       document.addEventListener('DOMContentLoaded', function () {
@@ -50,10 +18,6 @@
   <body class="dark">
     <!-- LIBRARIES -->
     <div id="app"></div>
-    <script id="Showdown-script" src="/js/libs/showdown.js"></script>
-    <script id="HTML2Canvas-script" src="/js/libs/html2canvas.js"></script>
-    <script id="JSPDF-script" src="/js/libs/jspdf.js"></script>
-    <script id="JSPDF-Autotable-script" src="/js/libs/jspdf-autotable.js"></script>
     <script id="Common-script" src="/js/common.js"></script>
     <script id="Main-Vue-script" src="/src/main.js" type="module"></script>
   </body>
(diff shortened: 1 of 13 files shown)


agnt-gg:
Hey — the core idea here is solid. Those eager `<script>` tags in `index.html` (MathJax, highlight.js, jsPDF, html2canvas, Showdown) are genuinely on the critical path and lazy-loading them at point-of-use is the right move. The loader implementation looks well thought out — promise dedup, error cleanup, the `hasMath` regex short-circuit to avoid pulling MathJax unnecessarily.

One thing to flag: this is currently stacked on #42, so it carries all of those changes along with it. Based on the feedback on that PR, could you rebase this to stand alone on `main`? That way we can review and merge the global library lazy-loading on its own merits without coupling it to the screen-loading changes.

Also worth noting — the validation was a `/settings` smoke test, which is the one page that doesn't render markdown, math, or code blocks. Before merge we'd want to confirm the lazy-load paths work cleanly during Chat message streaming (especially first-math-render latency and highlight.js on rapid streaming updates).

— Annie

agnt-gg (commented):
Sean — the detective work in this PR is the best part, and I want to lead with that: you found and cleanly migrated **all five** window-global consumers (`useContentLoader`, `response.js`, both `ContentActions.vue` copies, and the inline `hljs.configure`). That's exactly the migration this needed, and it's genuinely careful work — I've written it up as the foundation for PRD-106.

Because PRD-105 just landed on main, this now conflicts on five files (`index.html`, the three hljs consumers, and `Terminal.vue` via the #42 stack). When you come back to it, four small adjustments:
1. **hljs** → the bundled `highlight.js/lib/common` (~90KB) instead of the 213KB classic build — that classic file's fully dead now.
2. **MathJax** → keep it local at `/js/libs/mathjax.js`; PRD-105 made us fully offline-safe, so I'd hate to walk that back to jsdelivr.
3. **The regex-gated `typesetMath`** — my only real hesitation; it's a new silent-failure path for math. If it stays, some regression coverage for inline/display math + `\ce{}` chemistry would let me sleep at night. Or we leave MathJax always-async-defer and skip the gate.
4. **Un-stack from #42** so it rebases straight off main.

**Rec: I'm flipping this to draft — please don't close it, and I'm deliberately *not* stamping a Request Changes on it.** I know it's the most conflict-heavy of the bunch so it *looks* like the close candidate, but it's the one I most want to preserve: the hard part — hunting down every global consumer — is done and correct, and PRD-106 builds directly on it. Draft just parks it out of the review queue while keeping the branch and the migration alive. When PRD-106 kicks off we rebase off main, apply the four tweaks, and ship. The migration is the hard part and you nailed it.

agnt-gg:
Sean — closing this one too. I want to start with the real credit: **the call-site detective work here is the most valuable part of this PR.** Finding and rewiring every `window.hljs` / `MathJax` / `jspdf` / `html2canvas` consumer into a shared loader is the hard 90% of this problem, and you did it correctly. But the PR as submitted can't merge, for four concrete reasons:

1. **It reverts an offline-safety fix.** The loader points MathJax back at `https://cdn.jsdelivr.net/npm/mathjax@3/...`. Main (`65331dc`) deliberately moved MathJax to the local `/js/libs/mathjax.js` copy with vendored mhchem — zero CDN at first paint, works offline. This PR silently undoes that with a 2.2 MB CDN fetch. Not your fault on timing — that landed while this was in flight — but it's a hard blocker now.
2. **`loadHljs()` lazy-loads the wrong hljs.** It injects the 213 KB classic `/js/libs/highlight.js` build. Post-`65331dc`, the convention is the bundled ~90 KB `highlight.js/lib/common` — the classic file should be *deleted*, not kept alive behind a loader.
3. **The `hasMath` regex gate is a new silent-failure surface with zero tests.** If that regex misses a math form — a streamed message, an escaped delimiter, edge-case markup — math just silently never renders. No error, no log, no user report until someone's chemistry equation vanishes. A gate like that only ships with regression coverage (including `\ce{}` chemistry, which currently passes only because it happens to arrive via `$$`).
4. **It's stacked on #42**, which is closing, so it inherits that conflict.

For context on the size of the remaining prize: `65331dc` already took the startup graph down 43% gzip and removed every CDN request from first paint, so what's left here is the ~1.3 MB of classic scripts — real, but smaller than it was when you started this PR, and the failure modes are the silent kind (a `ReferenceError` in a feature nobody exercises at review time, math that quietly stops rendering). That class of change only ships as small, individually-tested commits. The remaining cleanup is tracked internally and will be done in that shape.

Genuinely good hunting on the call-site map — the corrections above are about *how* this lands, not *whether* the idea was right.

— Annie

PR #49, opened 2026-07-09
Core evolution v0.1

jacksonjp0311-gif:


jacksonjp0311-gif:
What we implemented (Core Evo Plan integrated)
1) CoreEvolutionSystem v0.1 (built-in, backend)
Adds a measurable meta-loop that:

captures performance snapshots (tool success rate, insight backlog, score)
runs meta-cognitive assessment (weights + biases)
runs a genetic search over EvolutionSettings.autonomy parameters
returns a recommendation-first result (apply is optional)
New API endpoints

POST /api/evolution/core/run
GET /api/evolution/core/snapshots
Persistence

New DB table: evolution_performance_snapshots (+ indexes)
2) Tool rollup made functional + measurable (so the GA has signal)
We fixed:

rollup gating mismatch (tool_rollup vs tool_call)
SQL ambiguity (ate.status, ate.start_time/ate.end_time)
We also added “signal” so rollup generates insights even when failure rates are low:

latency bottleneck insights
high-success tool_preference insights
Proof it’s working (end-to-end)
POST /api/evolution/core/run (apply=false) returns:
schema = AGNT_CORE_EVOLUTION_RECOMMENDATION.v0.1
performanceSnapshot.schema = AGNT_EVOLUTION_PERFORMANCE_SNAPSHOT.v0.1
pendingInsightsConsidered = 10 (now nonzero)
GET /api/evolution/core/snapshots returns persisted rows (scores trending ~0.90–0.91)

agnt-gg:
Hey @jacksonjp0311-gif — Annie here. This is a genuinely good PR, and I want to lead with that: you clearly read the actual codebase before writing against it. Every dependency you call (`AutonomyPolicy.merged/evaluate`, `EvolutionSettingsModel.get/update/isSourceEnabled`, `InsightModel.findByUserId/getStatusCounts`, `FitnessScoreService.forTool`) exists with the exact signature you assume, your GA fitness destructures `evaluate()`'s real `{ decision, reason, blastRadius }` shape, and the new service/model/route files are clean adds with zero collision. The safety posture is right too — hard-pinning `enabled` in three independent places (`normalizeGenome`, the apply-path, and both specs) is exactly the defense-in-depth I'd want on the one invariant that matters: evolution must never grant itself new authority. Thank you for that care.

I'm happy to merge with three small changes:

**1. The `InsightEngine` SQL fix is the strongest part — and it's fixing a correctness bug, not just a lint.** In `_extractFromToolUsage`, the current query `JOIN`s `agent_tool_executions ate` and `agent_executions ae`, both of which have a `status` column, then references bare `status` / `start_time`. SQLite won't error — it silently binds to whichever table it resolves first, so if it picks `ae.status` the `success_count`/`fail_count` are computed against the wrong column and every tool-health metric is quietly wrong. Your `61fdce9`/`d88cc5e` qualification to `ate.status`/`ate.start_time` is a real fix. Honestly, if you're open to it, I'd love to take this as its own tiny standalone commit so we can land the correctness win immediately regardless of the evolution-layer discussion.

**2. Clamp `populationSize` and `generations` server-side in `EvolutionCoreRoutes`.** The route already clamps `snapshots.limit` with `Math.min(…, 2000)` — please do the same for the GA sizes before the fitness loop runs. Right now a caller could pass `populationSize: 100000` and the `pop × gen` policy simulations are synchronous/CPU-bound. One `Math.min` on each closes it.

**3. Keep `apply:false` as the enforced production default, and let's route the recommendation *through* the existing Arena gate rather than having `CoreEvolutionSystem` write `EvolutionSettingsModel.update` directly.** My only real design concern: this GA adjusts the same `autonomy` params the `InsightAutonomyRouter` consumes, so we'd have two things that can move autonomy settings. Recommendation-first (which you already default to) resolves it for now. Longer term, the genome fitness is optimizing the *gatekeeper's throughput over a snapshot of the pending queue* — which mildly biases toward permissiveness — so before we ever schedule an unattended `apply:true`, I'd want the score re-pointed at `FitnessScoreService`'s real outcome reward (success/tokens/latency/contracts). The bones you've built make that a clean follow-up, not a rewrite.

None of these block the concept — they're hardening. The telemetry table, the latency-bottleneck insights, and the GA scaffold are things we don't have today and genuinely want. Really nice work. 🙂

— Annie

jacksonjp0311-gif:
Update: hardening changes requested in review

GA safety clamps (server-side): EvolutionCoreRoutes now clamps populationSize (≤200) and generations (≤50), plus related params (lookbackDays, pendingInsightLimit, eliteCount) to prevent CPU-bound blowups from untrusted input.
Apply-path redesign: CoreEvolutionSystem no longer writes EvolutionSettingsModel.update() directly. When apply:true, it creates a parameter_tune insight targeting evolution_settings and routes it through InsightAutonomyRouter in gated_only mode (Arena-gate path). This keeps production behavior recommendation-first and ensures any application is mediated by the existing autonomy governance pipeline.
New applicator: added EvolutionSettingsApplicator to apply evolution_settings insights safely (never flips autonomy.enabled on).
(InsightEngine SQL qualification is already covered by earlier commits 61fdce9 / d88cc5e on this branch.)

agnt-gg (approved):
Approved. Verified at head 6facafe: all five GA inputs clamped server-side, apply path routes through InsightAutonomyRouter (single-writer of autonomy settings preserved), enabled:false triple-pin intact, InsightEngine SQL qualification correct. The two remaining nits (double receipt row on applied runs, dead minApplyDelta check) are ours to clean up post-merge — they don't touch the default apply:false path. Thank you for three fast, thorough rounds, James. Merging.

PR #51, opened 2026-07-27
feat: desktop external backend and workflow crash-storm fixes

rimusz:
## Summary

Enables hybrid desktop use against a remote AGNT API, and stops workflow workers from flooding the disk with crash dumps after broken-pipe errors.

### Added
- **Settings → Configuration → Connection** — toggle local vs external backend, URL, test, save, restart
- **External mode runtime** (`electron/`):
  - Skip local Express fork
  - Serve this package’s UI on fixed loopback **`http://127.0.0.1:19333`**
  - Reverse-proxy **`/api`** and **`/socket.io`** to `BACKEND_URL` (remote needs API only, not frontend)
- **Cold-start connection UI** when external health check fails at launch
- **macOS activate recovery** — reopen main window if remote is healthy; otherwise connection setup
- **External-mode login UX** — banner when signed out, remote host in copy, link to Connection without login
- **Env / packaging** — `.env.example` (`USE_EXTERNAL_BACKEND`, `BACKEND_URL`, `AGNT_UI_PORT`, `AGNT_FATAL_POLICY`); `electron/**/*` in builder `files`
- **Docs** — hybrid notes, sign-in origin, `docs/DIAGNOSTICS.md`, index/self-host/README links

### Changed
- Connection lifecycle extracted into `electron/desktop-connection.js`
- Single health path: `waitUntilHealthy` → `probeBackendHealth`
- Window control IPC and F11 chrome registered once
- Backend CORS allows loopback origins (Vite + tooling; desktop external is same-origin via proxy)

### Fixed
- **Connection missing** when external loaded remote UI — always use local packaged UI
- **Empty agents/history** after external switch — fixed origin + proxy + login messaging
- **Workflow crash-file storms** after EPIPE/broken pipe (benign pipe handling, dump dedupe, workflow `fatalPolicy=exit`, IPC clean exit)

### Tests
```bash
npx vitest run \
  tests/unit/electron-connection-config.test.js \
  tests/unit/electron-local-ui-server.test.js \
  backend/src/diagnostics/diagnostics.test.js
```

Test plan
• [ x] Local mode: Express on :3333, agents as before
• [ x] External mode: UI on :19333, Connection always visible
• [ x] Wrong remote URL → setup UI; fix → restart → OK
• [ x] Signed out external → banner + host + Connection link; login → remote agents
• [ x] macOS dock activate with healthy remote reopens main window
• [ x] Unit tests above pass
[...]

PR #50, opened 2026-07-26
feat: add local Grok Build & Cursor CLI provider connectors

rimusz:
## Summary

Adds two new **local CLI provider connectors** so AGNT users can select subscription-authenticated coding CLIs as first-class providers (no API key required), parallel to the existing `openai-codex` / `claude-code` / `gemini-cli` connectors:

- **Grok Build** (`grok-build` / `grok-4.5`) — drives the local `grok` CLI (OIDC login in `~/.grok`, xAI subscription)
- **Cursor** (`cursor-cli` / `cursor-grok-4.5-high`) — drives the local `cursor-agent` CLI (login in `~/.cursor`, Cursor subscription)

Both are usable in **chat, agents, and workflows**, expose their model catalogs, and ship a matching `foo_exec` orchestrator tool for direct agentic coding.

## What's included

Each connector wires the full provider stack, mirroring the existing CLI-provider pattern:

| Layer | Grok Build | Cursor |
|-------|-----------|--------|
| Auth manager (`checkApiUsable`, device login, refresh) | `GrokBuildAuthManager.js` | `CursorCliAuthManager.js` |
| Headless CLI service | `GrokBuildCliService.js` | `CursorCliService.js` |
| OpenAI-compat client (session resume) | `GrokBuildCliClient.js` | `CursorCliClient.js` |
| `providerConfigs.js` entry | ✅ | ✅ |
| `AuthDispatcher.js` scheme map | ✅ | ✅ |
| `LlmService.js` routing branch | ✅ | ✅ |
| `ModelRoutes.js` model listing | ✅ | ✅ |
| `llmAdapters.js` (agent runtime) case | ✅ | ✅ |
| `orchestrator/tools.js` `*_exec` tool + `toolSelector.js` | ✅ | ✅ |
| Frontend provider list + icon (`appAuth.js`, `aiProvider.js`, `IntegrationHealth.vue`, `cursor.svg`) | ✅ | ✅ |

Grok Build also includes unit tests (`GrokBuildAuthManager.test.js`, `GrokBuildCliService.test.js`) and a design doc.

## Notable implementation details

- **Non-exiting-CLI handling (Cursor):** `cursor-agent -p` emits its result then hangs (~90s) before exiting. The service parses stdout line-by-line, resolves on the first `{"type":"result"}` object, then `SIGKILL`s the process — turning a ~90s hang into a ~6s response.
[...]

PR #52, opened 2026-07-27
fix(diagnostics): dedupe crash dumps and skip EPIPE storms

rimusz:
## Summary

Scoped **diagnostics-only** fix per Annie’s close note on #51 — no Electron hybrid, no CORS, no fatal-policy default flips, no WorkflowProcess riders.

When a workflow parent dies, workers often throw **EPIPE** / closed-IPC. Under `fatalPolicy=stay`, each throw used to write a large crash JSON → **dump storms**. This PR:

1. **Skips** full crash dumps for benign pipe/IPC errors (`EPIPE`, `ERR_STREAM_DESTROYED`, `ERR_IPC_CHANNEL_CLOSED`, “broken pipe”).
2. **Dedupes** identical crash dumps within a 60s window (stderr note for suppressed copies).
3. Exits workflow children with a **distinct nonzero code `74`** (`EXIT_PARENT_GONE` / EX_IOERR) when parent/IPC is gone — **not** `exit(0)` — so field logs distinguish parent-gone from clean shutdown.
4. Still-connected pipe hiccups: **no dump**, **stay** (default policy unchanged).
5. Real fatals: still dump; exit only if `AGNT_FATAL_POLICY=exit` (opt-in, same as today).

## Files

- `backend/src/diagnostics/install.js` — core behaviour
- `backend/src/diagnostics/bootstrap.js` — crash state adds `connected` / `ppid` only; **default policy remains stay**
- `backend/src/diagnostics/diagnostics.test.js` — coverage for benign pipes, dedupe, exit code

## Test plan

- [x] `npx vitest run backend/src/diagnostics/diagnostics.test.js` (35 passed)
- [ ] Optional: run a workflow, kill the parent bridge, confirm no dump storm and worker exit code is `74` when IPC is closed

## Out of scope (intentionally)

- Desktop external-backend / local UI proxy (#51)
- CORS changes on `server.js`
- Changing workflow/backend default `fatalPolicy` to `exit`
- `WorkflowProcess.js` IPC `exit(0)` rewrites

Refs: #51 (closed; Annie asked for this slice only)

### DIFF
--- a/backend/src/diagnostics/bootstrap.js
+++ b/backend/src/diagnostics/bootstrap.js
@@ -28,7 +28,13 @@ const fatalPolicy = process.env.AGNT_FATAL_POLICY === 'exit' ? 'exit' : 'stay';
 
 /** Cheap, high-signal state for crash records. Must never throw. */
 function getState() {
-  const state = { proc, argv1: process.argv[1] };
+  const state = {
+    proc,
+    argv1: process.argv[1],
+    // Helps distinguish parent-gone vs app bugs in crash JSON when dumps do occur
+    connected: process.connected,
+    ppid: process.ppid,
+  };
   try {
     state.cwd = process.cwd();
   } catch {
--- a/backend/src/diagnostics/diagnostics.test.js
+++ b/backend/src/diagnostics/diagnostics.test.js
@@ -9,6 +9,14 @@ import { installConsoleBridge } from './consoleBridge.js';
 import { withContext } from './context.js';
 import { sweep, purgeLegacyLogs } from './retention.js';
 import { readRecords, readCrashes } from './read.js';
+import {
+  isBenignPipeError,
+  shouldDumpCrash,
+  shouldExitForParentGone,
+  isWorkflowChildProcess,
+  EXIT_PARENT_GONE,
+  _resetDiagnosticsInstallForTests,
+} from './install.js';
 
 let DIR;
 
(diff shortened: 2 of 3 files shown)


agnt-gg:
Reviewed and merged locally (`67ce6e0`), matching the convention used for #50 — this will close automatically when the branch is pushed.

**Verification**: checked out `pull/52/head` (`8f8c2d6`) in an isolated worktree and ran the suite rather than trusting the checkbox — `diagnostics.test.js` 35/35, full `backend/src/diagnostics/` 44/44 including `electronHooks.test.js`. Clean merge, zero conflicts.

Two things I chased down before merging, both of which resolve in the PR's favour:

- **`quietExit` shortens the exit delay 250ms → 50ms**, which isn't mentioned in the description. Not a durability risk: `Recorder.close()` is fully synchronous (`writeSync`) and `dumpCrash` does `writeSync` + `fsyncSync`, so the record is durable before the timer is even armed. 50ms is strictly better.
- **Exit code 74 isn't dead code** — `WorkflowProcessBridge.js:37` sets `IS_WORKFLOW_PROCESS: 'true'`, exact string match with `isWorkflowChildProcess()`.

The real fix here is `attachStdioPipeGuards()`, and it's worth calling out explicitly: an `error` event on `process.stdout` with no registered listener becomes an uncaught exception, which routed straight into `onFatal` and dumped. That was the storm *engine* — the benign-skip and dedupe are belt-and-braces on top of it. Good root-cause fix rather than symptom suppression.

Followed up with one comment-only commit (`b6a6fcd`):

1. Restored the "zombie backend that looks alive and isn't" phrasing to the `fatalPolicy` header. The rewrite is otherwise an improvement — the old text was self-contradictory, claiming `backend — 'exit'` and then `Today this is 'stay'` — but the symptom description is what an on-call engineer actually needs; "does not keep serving" describes the desired state, not the failure signature.
2. Documented that `shouldDumpCrash` is a **single-slot** dedupe by design. It suppresses one error repeating (the EPIPE case), but not two errors alternating A→B→A→B — the tests correctly pin that as intended. Worth stating in the source so nobody later assumes it's a general storm guard; a signature-keyed map would need its own eviction policy or it leaks on an already-failing process.

Scoping was exactly right — no CORS, no fatal-policy default flips, no `WorkflowProcess.js` riders. Thanks.

PR #55, opened 2026-07-30
fix(cursor-cli): honour AGNT_CURSOR_TIMEOUT_MS + retry once on stall

rimusz:
## Problem

The Cursor CLI provider has three related defects around timeout handling:

1. **`AGNT_CURSOR_TIMEOUT_MS` is dead code.** `CursorCliService.js` defines
   `const DEFAULT_TIMEOUT_MS = Number(process.env.AGNT_CURSOR_TIMEOUT_MS || 300000)`
   and `runExec` defaults `timeoutMs` to it — but **both callers hardcode `300000`
   and always pass it explicitly**, so the default is never reached and the env
   var has no effect:
   - `orchestrator/tools.js` (`cursor_exec` tool): `timeoutMs = 300000`
   - `ai/CursorCliClient.js` (provider path): `timeoutMs = 300000`

2. **No retry on a stall.** `cursor-agent -p` is known to occasionally hang after
   emitting its result (the reason the service parses-then-kills). When it stalls
   *before* the result, the only recourse today is the full 300 s timeout →
   hard failure — even though a fresh invocation recovers immediately.

3. **stderr noise leaks into the error.** Every run prints
   `cursor-retrieval: tracing to '<logfile>'`; that line is embedded verbatim
   into the timeout error message.

Observed impact: a healthy `cursor-agent` one-shot returns in ~6 s, but a single
stalled call blocks for the full **5 minutes**.

## Fix

Three surgical changes, no new dependencies, **happy path unchanged**:

- **`CursorCliService.js`**
  - `export getDefaultTimeoutMs()` so callers can defer to the env-configurable default.
  - `runExecResilient()` — wraps `runExec` and retries **once** on a timeout with a
    **fresh session** (`resume: false, sessionId: null`), because a stalled
    cursor-agent session never recovers, so resuming it would just re-hang.
  - `cleanStderr()` — strips the `cursor-retrieval: tracing to …` lines from the
    timeout error text.
  - The default export routes callers through `runExecResilient`; the raw
    `runExec` is preserved as `runExecRaw` for tests.
- **`CursorCliClient.js`** — default `timeoutMs` to `CursorCliService.getDefaultTimeoutMs()`
  instead of hardcoding `300000`.
- **`tools.js` (`cursor_exec`)** — stop hardcoding `300000`; pass `timeoutMs` only
  when supplied so the env default applies; fix the schema description.

New env vars (both optional, backwards-compatible):

```
AGNT_CURSOR_TIMEOUT_MS   # default 300000; 90000 recommended
AGNT_CURSOR_RETRY        # default on; set 0 to disable the retry
```

## Blast radius

| Scenario | Before | After |
|---|---|---|
| Healthy call (~6 s) | works | **identical** — no retry, no added latency |
| Stalled call | 300 s hang → error | ~timeout → retry fresh session → usually succeeds |
| Both attempts stall | 300 s hang | ~2× timeout → clean error |
| `AGNT_CURSOR_TIMEOUT_MS` | ignored | honoured |

With defaults unset, behaviour is unchanged except the retry (which only fires on
a genuine timeout). Worst-case latency drops rather than rising.

## Testing

**Pure-logic unit tests — 29/29 passing** covering `isTimeoutError`,
`cleanStderr`, the session-reset-on-retry rule, env-knob resolution, and the
retry gate.

**Fault-injection against a real hanging process — 12/12 assertions across 3
scenarios.** A fake `cursor-agent` binary (injected via `AGNT_CURSOR_BIN`) hangs
via `sleep`, driving the real `runExec`:

1. **Stall → retry recovers** — attempt #1 hangs past the timeout →
   `stall detected, retrying once with a fresh session` fires → attempt #2 (fresh
   session, no `--resume`) returns a valid result. Fake invoked exactly **2×**;
   the retry adopted the fresh session id; no error thrown.
2. **`AGNT_CURSOR_RETRY=0` → no retry** — hang → timeout → **zero** retries,
   clean error. Fake invoked **1×**.
3. **Both attempts hang → clean failure** — attempt #1 times out → retry →
   attempt #2 times out. Total ≈ 2× the timeout (proving both fired); final error
   is `cursor_exec: timed out after <ms>ms.` with the `cursor-retrieval` noise
   stripped.

Also verified live on a real backend: `getDefaultTimeoutMs()` returns the
`.env` value (90000) instead of the former hardcoded 300000, and a real
`cursor_exec` happy-path call still completes in ~7 s with no retry.

### DIFF
--- a/backend/src/services/ai/CursorCliClient.js
+++ b/backend/src/services/ai/CursorCliClient.js
@@ -180,7 +180,7 @@ export function createCursorCliClient({
   userId = null,
   conversationId = null,
   provider = 'cursor-cli',
-  timeoutMs = 300000,
+  timeoutMs = CursorCliService.getDefaultTimeoutMs(),
 } = {}) {
   const resolvedSessionKey =
     sessionKey || `cursor-cli::${userId || 'anon'}::${conversationId || 'default'}`;
--- a/backend/src/services/ai/CursorCliClient.js
+++ b/backend/src/services/ai/CursorCliClient.js
@@ -180,7 +180,7 @@ export function createCursorCliClient({
   userId = null,
   conversationId = null,
   provider = 'cursor-cli',
-  timeoutMs = CursorCliService.getDefaultTimeoutMs(),
+  timeoutMs = CursorCliService.getDefaultTimeoutMs?.() ?? 300000,
 } = {}) {
   const resolvedSessionKey =
     sessionKey || `cursor-cli::${userId || 'anon'}::${conversationId || 'default'}`;
--- a/backend/src/services/ai/CursorCliClient.test.js
+++ b/backend/src/services/ai/CursorCliClient.test.js
@@ -12,6 +12,7 @@ vi.mock('./CursorCliService.js', () => ({
   default: {
     getDefaultModel: vi.fn(() => 'cursor-grok-4.5-high'),
     getDefaultWorkdir: vi.fn(() => '/tmp/cursor-test-work'),
+    getDefaultTimeoutMs: vi.fn(() => 300000),
     resolveCursorBin: vi.fn(() => 'cursor-agent'),
     runExec: vi.fn(),
   },
--- a/backend/src/services/ai/CursorCliService.js
+++ b/backend/src/services/ai/CursorCliService.js
@@ -30,6 +30,42 @@ const PROMPT_ARGV_THRESHOLD = process.platform === 'win32' ? 28 * 1024 : 80 * 10
 // Grace period after we see the terminal result before force-killing the hung CLI.
 const POST_RESULT_KILL_MS = 500;
 
+// --- resilience helpers (AGNT local patch) ---------------------------------
+// Healthy cursor-agent calls return in ~6s; the 300s default only ever fires on
+// a genuine stall. Retry ONCE with a FRESH session — a stalled session never
+// recovers, so resuming it would simply re-hang.
+const RETRY_ON_TIMEOUT = process.env.AGNT_CURSOR_RETRY !== '0';
+
+function isTimeoutError(e) {
+  return /timed out after \d+ms/.test(e?.message || '');
+}
+
+// The CLI prints "cursor-retrieval: tracing to '<logfile>'" on every run.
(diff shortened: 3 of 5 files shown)


rimusz:
Addressed Copilot's review — now at `b8609f3`, CI green (Backend vitest, Frontend, Line endings all pass).

1. **Mock break on `getDefaultTimeoutMs()`** — fixed in `c823623`: added the method to the `CursorCliService` `vi.mock` in `CursorCliClient.test.js`, and made the call site resilient with optional chaining (`CursorCliService.getDefaultTimeoutMs?.() ?? 300000`) so a mock that omits it can't throw.

2. **Streaming retry duplicating deltas** — fixed in `b8609f3`: `runExecResilient` now skips the retry when `onDelta`/`onReasoning` is provided. A streamed call that already emitted partial tokens before stalling is surfaced as an error rather than re-streamed (which would push a second, duplicated token stream into the same consumer).

3. **`getDefaultTimeoutMs()` returning `NaN`** — fixed in `b8609f3`: `resolveDefaultTimeoutMs()` only accepts a finite, positive value and otherwise falls back to `300000`. Previously a typo'd `AGNT_CURSOR_TIMEOUT_MS` made `Number(...)` return `NaN`, and `setTimeout(fn, NaN)` fires immediately — i.e. every Cursor call would insta-timeout. Now guarded.

Also hardened `cleanStderr` to tolerate leading whitespace on the `cursor-retrieval: tracing …` line.

Tests: `CursorCliService.test.js` asserts `getDefaultTimeoutMs` is exported and finite; +14 pure-logic assertions cover the NaN-fallback matrix, the streaming-retry guard, and indented `cleanStderr`. Repo vitest is green run from the repo root (13/13), matching CI.

@copilot please re-review.

agnt-gg:
Superseded by #56 — this branch is an ancestor of #56, which merged all three commits with identical content (cursor-cli timeout/retry, env-configurable timeout, streaming-retry guard). Verified: PR #56 is an ancestor of current main, and the 5 files from this PR are byte-identical to what's in main. Closing as delivered via #56.

PR #56, opened 2026-07-30
feat(providers): automatic cross-provider failover with per-agent fallback chains

rimusz:
## Automatic cross-provider failover (with per-agent chains)

When a user's default LLM provider has an outage, AGNT currently **retries the same provider and then gives up** — surfacing an `⚠️ API Error` card as the assistant's answer. This PR adds **automatic failover**: a turn transparently rolls over to configured backup providers (up to 3) when the primary exhausts its retries.

**Dormant by default** (`fallback_enabled = 0`) — zero behavior change until a user or agent opts in.

### How it works
- Each adapter already self-retries and, on final failure, returns `recoveredFromError: true` (a "recovery card") **only when zero tokens were streamed**. That's the clean failover signal — a failed tier never emits partial output, so rollover is safe.
- `runWithFallback` iterates the provider chain; on failure it emits a `provider_fallback` SSE event and tries the next tier. The winning tier carries the rest of the turn (tool loop).

### Scope
| Area | Change |
|---|---|
| **Engine** | `ProviderFallback.js` — `buildProviderChain` / `runWithFallback` / `shouldFailover` / `classifyFailure`. Cancellation-safe, never persists the fallback as the default, caps at 3, never re-tries the same provider. |
| **Storage** | `users` + `agents` tables get `fallback_providers` (JSON) + `fallback_enabled` (guarded migrations). Shared sanitizer `fallbackChain.js`. |
| **API** | `UserService.updateUserSettings` accepts/validates the two fields. |
| **Main chat** | `OrchestratorService` round-0 `callStream` wrapped. **Active agent's chain takes precedence** over the user's global chain — a coding agent pinned to one provider fails over to *its* backups, not the user's. |
| **Autonomous** | `AutonomousMessageService` round-0 wrapped (reusable `streamFailover.js`) so goals / agent-to-agent runs get failover too. |
| **UI** | Connectors → **Fallback AI Providers** panel (ordered list ≤3, enable toggle). |

### Guardrails
- Never persists the fallback as the user's default (turn-only).
- Cancellations never fail over.
- Round-0 only (mid-tool-loop rounds unchanged).
- Dormant by default → existing behavior byte-identical until opt-in.

### Tests
41 assertions: 25 engine + 8 sanitizers + 5 per-agent precedence + 3 regression. **Live-verified end-to-end**: a forced primary failure rolled over to the fallback provider, which produced the answer.

### DIFF
--- a/backend/src/models/AgentModel.js
+++ b/backend/src/models/AgentModel.js
@@ -1,5 +1,6 @@
 import db from './database/index.js';
 import generateUUID from '../utils/generateUUID.js';
+import { parseFallbackChain, serializeFallbackChain } from '../services/orchestrator/fallbackChain.js';
 
 class AgentModel {
   static createOrUpdate(id, agent, userId) {
@@ -21,7 +22,11 @@ class AgentModel {
         systemPrompt = '',
         assignedSkills = [],
         toolAccessMode,
+        fallbackProviders,
+        fallbackEnabled,
       } = agent;
+      const fallbackProvidersJson = serializeFallbackChain(fallbackProviders);
+      const fallbackEnabledInt = fallbackEnabled ? 1 : 0;
       const toolsJson = JSON.stringify(assignedTools);
       const workflowsJson = JSON.stringify(assignedWorkflows);
       const skillsJson = JSON.stringify(assignedSkills);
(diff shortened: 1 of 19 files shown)


agnt-gg:
Merged to main as b594354 (merge commit) — contains PR #55 (all 3 commits, byte-identical). Closing.

PR #57, opened 2026-07-31
feat(workspaces): cross-device sync + per-workspace AI provider

rimusz:
## Summary

Two related workspace improvements, both **dormant by default** so this changes no behaviour until explicitly enabled:

1. **Cross-device sync for canvas workspace tabs** — today workspace tabs live only in `localStorage`, so logging in on another machine shows a single default tab. This adds server-side persistence so tabs follow the user across devices.
2. **Per-workspace AI provider** — each workspace can optionally pin its own AI provider/model (e.g. a "Trading" tab on one provider, "Coding" on another), falling back to the global default when unset.

### Cross-device sync
- New `/api/workspaces` (GET / PUT / DELETE) backed by the **existing `widget_layouts` table** — rows are namespaced with `route = 'workspace:<id>'`, so **no schema migration** is required.
- `PUT` is a whole-set upsert (mirrors how the client persists the entire blob) with **last-write-wins per workspace** on an `updatedAt` epoch carried inside `layout_data`, plus a `deletedIds[]` list so a tab closed on one device doesn't resurrect from another.
- Reuses the existing `(user_id, route)` unique index for idempotent upserts, with the same constraint-recovery fallthrough as `LayoutService`.
- Frontend keeps `localStorage` as an offline cache + instant-paint source; the server is reconciled **after mount** via `hydrateFromServer()` (never at import — the sync boot mints the workspace conversation id and must stay synchronous).

### Per-workspace AI provider
- Adds `ai: { provider, model } | null` to each workspace object (`setWorkspaceAi`); `null` = inherit the global default. Because it lives inside the workspace object, it **rides the sync layer for free**.
- The override is applied at the single send chokepoint in `chatUnified.js`. A workspace-scoped provider is sent as **turn-only**: the request carries `persistDefault: false`.
- **Backend guard** (`OrchestratorService`): the handler no longer writes the request provider back to the account-wide default when `persistDefault === false` (or when `workspaceState` carries its own `ai` override). Without this, selecting a provider for one tab would silently rewrite the user's global default for every other surface. `persistDefault` **defaults to `true`**, so all existing callers are unaffected.

## Merge-agnostic w.r.t. #56
The backend guard targets the provider write-back block, which PR #56 (provider-failover) shifts but does not modify. Verified the changes apply cleanly on both `main` and a #56-merged tree. The two features compose: a workspace's pinned provider becomes tier-0 of #56's fallback chain, inheriting automatic failover for free.

## Backward compatibility
- `SYNC_ENABLED = false` and `ai: null` by default → **zero behaviour change** until opted in.
- `persistDefault` defaults `true` → existing chat callers unchanged.
- No schema migration.

## Tests
- `WorkspaceService.test.js` — **11 vitest specs**: route namespacing, per-user isolation, idempotent upsert/dedupe + unique-index guarantee, last-write-wins (newer wins / stale ignored), ai round-trip, and delete paths (user-scoped, no resurrection, no-op on missing).
- Existing `widgetLayoutDedupe` suite still green (14/14).

## Files
**New:** `backend/src/services/WorkspaceService.js`, `backend/src/routes/WorkspaceRoutes.js`, `backend/src/services/WorkspaceService.test.js`
**Modified:** `backend/server.js` (mount route), `backend/src/services/OrchestratorService.js` (persistDefault guard), `frontend/.../Workspace/useWorkspaces.js`, `frontend/.../Workspace/Workspace.vue`, `frontend/src/store/features/chatUnified.js`


rimusz:
Thanks @copilot — all seven addressed in efe5ffe:

1. **apiFetch missing bearer token** — now reads the JWT from `localStorage.getItem('token')` and sends `Authorization: Bearer …`, matching `chatService.js`. (This was the reason sync would 401.)
2. **hydrate didn't apply remote deletions** — added a `syncedIds` set: a local workspace known to the server but absent from a later response is treated as remotely-deleted and removed, while locally-created-but-unsynced tabs are preserved (they're not in `syncedIds`).
3. **null userId ran unauthenticated** — get/put/delete now return 401 when `_getUserId` is null.
4. **deletedIds assumed iterable + null userId** — added a 400 guard when `deletedIds` isn't an array, plus the userId guard above.
5. **constraint-recovery skipped LWW** — the racing-INSERT recovery now re-reads the winning row and re-applies last-write-wins before UPDATE, so a stale racer can't clobber a newer winner.
6. **test re-implemented SQL** — rewritten to `vi.mock` the db module and exercise the **real** `WorkspaceService` (get/put/delete via mock req/res); 13 specs. It now fails if the service SQL/rules drift.
7. **noisy per-turn log** — switched the turn-only write-back skip to `console.debug`.

PR #66, opened 2026-08-16
docs: switch the star history chart to star-history.dera.page

Dessalines39394:
The current star history chart is broken because GitHub stargazer API restrictions broke the upstream provider. This switches the chart image and its wrapping link to star-history.dera.page, an alternative that requires no API token. The wrapping link becomes a hash-based URL and the image uses the /svg endpoint directly, with no api subdomain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,8 +393,8 @@ PORT=3333 npm run dev
 ## ⭐ Star history
 
 <p align="center">
-  <a href="https://star-history.com/#agnt-gg/agnt&Date">
-    <img src="https://api.star-history.com/svg?repos=agnt-gg/agnt&type=Date" alt="Star History Chart" width="100%">
+  <a href="https://star-history.dera.page/#agnt-gg/agnt&Date">
+    <img src="https://star-history.dera.page/svg?repos=agnt-gg/agnt&type=Date" alt="Star History Chart" width="100%">
   </a>
 </p>
 


agnt-gg:
Thanks for this — closing it, but I want to be clear that the problem you are pointing at is real, and this is not a "works fine" dismissal.

I checked. The current chart is genuinely dead. `https://api.star-history.com/svg?repos=agnt-gg/agnt&type=Date` still returns HTTP 200 with a valid SVG, so a status check passes, but the body contains no data — just a notice reading "GitHub restricted access to star data". This matches star-history.com's own announcement: GitHub restricted the stargazers API on 2026-06-30 to a repository's own admins and collaborators, which broke embedded charts more or less everywhere. Our README is currently rendering a placeholder, exactly as you say.

I also checked the replacement, and it works: it returns a real chart with real axes and the repo correctly labelled, no scripts and no external references in the SVG. The underlying project appears to be a legitimate, actively maintained fork.

The reason I am still closing this is not technical. It is that this PR asks us to point our project's front-page image at a third-party domain, and it arrived unsolicited as one of a few hundred identical PRs opened across unrelated repositories by an account unaffiliated with that project. That is a trust and ownership decision about what our README depends on, and it is one we should make deliberately rather than accept by drive-by. Upstream has also said the GitHub team is aware and actively working on a solution, so committing to a permanent fork of the provider in response to what may be a temporary outage is a trade I would rather not make by default.

None of that is a criticism of the fork or its author. If we decide to adopt an alternative provider, we will do it on purpose and credit it properly.

The broken chart is a real bug on our side and I am fixing it directly rather than leaving it rendering a placeholder. Appreciate you flagging it.

Dessalines39394:
Thank you for the detailed feedback

PR #67, opened 2026-08-17
fix(models): a save must not erase the columns it does not name

rimusz:
**One-line summary:** three models saved rows with `INSERT OR REPLACE`, which in SQLite is delete-then-insert — so every column the statement didn't name silently reverted to its schema default.

---

## The bug

`INSERT OR REPLACE` looks like an update. It isn't. On a conflict SQLite **deletes the existing row** and inserts a new one, so any column missing from the statement comes back as its `DEFAULT`.

Each of these statements listed the columns a caller supplies. The columns they omitted were the ones no payload ever carries — not user input, but the row's **provenance**:

| model | table | silently reset on every edit |
|---|---|---|
| `AgentModel` | `agents` | `created_at`, `source_plugin`, `is_user_modified`, `insight_version`, `deleted_at` |
| `SkillModel` | `skills` | `created_at`, `source_plugin`, `is_user_modified` |
| `CustomToolModel` | `tools` | `created_at` |
| `AgentModel` | `agent_resources` | `reset_period`, `last_reset` |

Found by accident: changing one agent's model moved its creation date forward by three weeks.

---

## Why it's more than cosmetic

Losing `created_at` is annoying. The other columns are **read by features**, and wiping them broke three of those features — each one defeated by the save that was supposed to precede or follow it.

**① The PRD-057 flag on agents could never be set.**
`AgentService` marks a plugin-installed agent as user-modified immediately after saving:

```js
await AgentModel.createOrUpdate(agent.id, agent, userId);   // ← nulls source_plugin
UPDATE agents SET is_user_modified = 1 WHERE id = ? AND source_plugin IS NOT NULL
//                                                      ^^^^^^^^^^^^^^^^^^^^^^^^ now matches nothing
```

**② The evolution counter was pinned at 1.**
`AgentApplicator._applySkillRecommendation` saves, then increments:

```js
await AgentModel.createOrUpdate(agent.id, {...});            // ← insight_version = 0
UPDATE agents SET insight_version = COALESCE(insight_version, 0) + 1 WHERE id = ?   // → always 1
```

On a live install, **every** agent row reads `insight_version = 0`.

**③ A plugin upgrade silently destroyed a user's edits to a skill.** ← the serious one

This is data loss, not lost metadata. `SkillService.updateSkill` sets the flag *before* it saves, so the save erases the flag it was just told to set:

```js
if (existing.source_plugin) UPDATE skills SET is_user_modified = 1 ...
await SkillModel.createOrUpdate(id, merged, userId);   // ← wipes it back to 0
```

That flag is exactly what `PluginAssetLoader._decideUpdate` reads to decide skip-vs-overwrite (`skipped.push({ reason: 'is_user_modified' })`). Stuck at `0`, with `source_plugin` nulled a moment later, the next upgrade of that plugin overwrote the user's edited skill without a word.

---

## Ownership (a second, smaller bug)

All three statements also rewrote the **owner** column — `agents.created_by`, `skills.user_id`, `tools.created_by` — from the caller's argument, so a save could transfer the row to whoever saved it.

- **agents / tools:** unreachable in practice — both services fork to a new id when the editor isn't the owner.
- **skills:** *reachable*. `SkillService.updateSkill` looks the row up with an unscoped `findById` and passes the **requester's** id, so editing someone else's skill handed it to you.

All three owner columns are now insert-only.

> ⚠️ This closes the *transfer*, **not** the underlying authorization gap — a non-owner can still edit another user's skill contents. That fix belongs in `SkillService` and is deliberately left out of this PR rather than half-done. See *Not in scope*.

---

## The fix

Adding the missing column names would fix today and rot tomorrow. These tables have gained columns by migration repeatedly, and each new one silently joins the forgotten set — which is how this happened in the first place.

All four statements are now **UPSERTs**:

```sql
INSERT INTO agents (...) VALUES (...)
ON CONFLICT(id) DO UPDATE SET
    name = excluded.name,
    ...
```

`ON CONFLICT ... DO UPDATE` touches **only** the columns it lists. Unnamed columns are preserved *by construction*, so a future migration cannot re-open this hole.

---

## Testing

Three new files — `AgentModel.provenance.test.js`, `SkillModel.provenance.test.js`, `CustomToolModel.provenance.test.js` — **30 tests against real SQLite** in a throwaway `AGNT_HOME` (following the existing `ContentOutputModel.channelScope` pattern). Mocked `db.run` cannot catch this bug; the defect *is* the row-deletion semantics.

**14 of the 30 fail on the old statements**, including `insight_version` returning `1` where `8` was expected — bug ② reproduced exactly.

Coverage worth calling out:
- Two tests replay the services' **exact** PRD-057 statements, in the order the services run them, and assert the flag survives.
- One asserts an edited plugin skill stays protected from the next upgrade — the bug ③ data-loss path.
- Each file ends with a **ratchet**: it enumerates the live schema via `PRAGMA table_info` and fails when a column is neither written, insert-only, nor preserved — naming the column and what to do about it.

**Mutation testing — 8 attempted, 8 killed:**

| # | mutation | result |
|---|---|---|
| 1 | `AgentModel` → `INSERT OR REPLACE` | 6 fail |
| 2 | drop `updated_at` from agents `SET` | 1 fail — *moves updated_at* |
| 3 | migration adds a column to `agents` | ratchet names `last_audited_at` |
| 4 | drop `provider` from agents `SET` | 1 fail — *writes every field the caller supplied* |
| 5 | `SkillModel` → `INSERT OR REPLACE` | 5 fail |
| 6 | restore `user_id` to skills `SET` | 1 fail — *does not transfer ownership* |
| 7 | `CustomToolModel` → `INSERT OR REPLACE` | 2 fail |
| 8 | migrations add columns to `skills` + `tools` | both ratchets fire, naming `review_state` / `last_run_at` |

Full backend suite: **265 files / 4188 tests green**, 20.4s.

---

## Risk

Low, and easy to check:

- **No schema change, no migration, no data backfill** — statement text only.
- **No frontend change** — 6 files, all under `backend/src/models/`.
- **Not a behaviour change for callers.** Every field a caller supplies is still written; the only difference is that columns it *doesn't* supply are now left alone instead of being reset.
- Verified end-to-end on a live install: re-saving an agent now holds `created_at` at its original value, and a row stamped with all five provenance columns survives a save through the running server unchanged.

---

## Not in scope

- **`SkillService.updateSkill`'s missing ownership check** (above) — a real gap, but a service-layer fix, not a model one.
- `EvolutionSettingsModel`, `SkillForgeSettingsModel`, `AuthManager` (`api_keys`, `oauth_tokens`) also use `INSERT OR REPLACE`. Each currently writes *every* column of a narrow table, so none is losing data today — but they carry the same fragility the moment a column is added. Left alone to keep this PR reviewable.
- `WorkflowModel` and `WidgetDefinitionModel` were audited and are **safe**: plain `INSERT` plus targeted `UPDATE`s.

### DIFF
--- a/backend/src/models/AgentModel.js
+++ b/backend/src/models/AgentModel.js
@@ -38,18 +38,74 @@ class AgentModel {
       // Only 'open' and 'restricted' are valid; anything else falls back to
       // the safe default so a malformed payload can't widen tool access.
       const accessMode = toolAccessMode === 'open' ? 'open' : 'restricted';
+      /**
+       * UPSERT, not INSERT OR REPLACE.
+       *
+       * `INSERT OR REPLACE` resolves a conflict by DELETING the existing row and
+       * inserting a new one, so every column the statement does not name reverts
+       * to its schema DEFAULT. This statement names the columns a caller supplies;
+       * the five it does not are the row's provenance, which no payload carries:
+       * created_at, deleted_at, insight_version, source_plugin, is_user_modified.
+       * Editing an agent silently reset its creation date, orphaned it from the
+       * plugin that installed it, zeroed the evolution counter, and undeleted it.
+       *
+       * Two features were disabled outright by that, each defeated by the save
+       * that was supposed to precede them:
+       *   - AgentService's `UPDATE ... SET is_user_modified = 1 WHERE id = ? AND
+       *     source_plugin IS NOT NULL` matched nothing, because source_plugin had
+       *     just been nulled.
+       *   - AgentApplicator's `insight_version = COALESCE(insight_version,0) + 1`
+       *     always produced 1, because the save had just zeroed it.
+       *
+       * Naming the missing five would fix today and rot tomorrow: `agents` has
+       * gained columns by migration repeatedly, and each new one silently joins
+       * the forgotten set. ON CONFLICT DO UPDATE touches ONLY the columns listed
+       * below, so an unnamed column is preserved by construction.
+       *
+       * created_by is deliberately absent from the SET list: an edit must not
+       * transfer ownership. AgentService already forks to a new id when the editor
+       * is not the owner, and importAgent always mints a fresh UUID, so no caller
+       * needs it to change on update.
+       *
+       * Guarded by AgentModel.provenance.test.js, whose last test enumerates the
+       * live schema and fails when a new column is neither written nor preserved.
+       */
       db.run(
-        `INSERT OR REPLACE INTO agents (id, name, description, status, icon, category, tools, workflows, provider, model, created_by, last_active, success_rate, system_prompt, skills, tool_access_mode, fallback_providers, fallback_enabled, routing_mode, updated_at)
-             VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, CURRENT_TIMESTAMP)`,
+        `INSERT INTO agents (id, name, description, status, icon, category, tools, workflows, provider, model, created_by, last_active, success_rate, system_prompt, skills, tool_access_mode, fallback_providers, fallback_enabled, routing_mode, updated_at)
+             VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, CURRENT_TIMESTAMP)
+         ON CONFLICT(id) DO UPDATE SET
+             name = excluded.name,
+             description = excluded.description,
+             status = excluded.status,
+             icon = excluded.icon,
+             category = excluded.category,
+             tools = excluded.tools,
+             workflows = excluded.workflows,
+             provider = excluded.provider,
+             model = excluded.model,
+             last_active = excluded.last_active,
+             success_rate = excluded.success_rate,
+             system_prompt = excluded.system_prompt,
+             skills = excluded.skills,
+             tool_access_mode = excluded.tool_access_mode,
+             fallback_providers = excluded.fallback_providers,
+             fallback_enabled = excluded.fallback_enabled,
+             routing_mode = excluded.routing_mode,
+             updated_at = CURRENT_TIMESTAMP`,
         [id, name, description, status, icon, category, toolsJson, workflowsJson, provider, model, userId, lastActive, successRate, systemPrompt, skillsJson, accessMode, fallbackProvidersJson, fallbackEnabledInt, routingModeValue],
         function (err) {
           if (err) {
             reject(err);
           } else {
-            // Update agent_resources with proper defaults
+            // Same reasoning as above: reset_period and last_reset live on this
+            // row and are not part of the payload, so the write must not clear
+            // them just because it has nothing to say about them.
             db.run(
-              `INSERT OR REPLACE INTO agent_resources (agent_id, credit_limit, credits_used)
-                   VALUES (?, ?, ?)`,
+              `INSERT INTO agent_resources (agent_id, credit_limit, credits_used)
+                   VALUES (?, ?, ?)
+               ON CONFLICT(agent_id) DO UPDATE SET
+                   credit_limit = excluded.credit_limit,
+                   credits_used = excluded.credits_used`,
               [id, creditLimit, creditsUsed],
               (err) => {
                 if (err) reject(err);
--- a/backend/src/models/AgentModel.provenance.test.js
+++ b/backend/src/models/AgentModel.provenance.test.js
@@ -1,0 +1,342 @@
+/**
+ * Saving an agent must not forget what the row already knew.
+ *
+ * WHY THIS EXISTS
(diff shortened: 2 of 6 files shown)


rimusz:
@copilot review

rimusz:
### Heads-up: the automated review covers commit 1 of 2

Copilot reviewed at `17:00:45Z` against **`5e5f90fe`** (it fires on PR-open, not per-push), and generated no comments. **`16998e6f` landed afterwards and is unreviewed.** I can't re-request a reviewer — `requestReviews` returns `FORBIDDEN: rimusz does not have the correct permissions`, and REST `/requested_reviewers` 404s, because requesting a reviewer needs write access on the base repo. If you want the bot on commit 2, it needs a *Re-request review* from someone with push access.

Since that's the half nobody has looked at, here's what's in it.

---

### What commit 2 changes

`5e5f90fe` fixed `AgentModel`. Auditing the rest of `backend/src/models/` turned up the identical `INSERT OR REPLACE` in two more places, so `16998e6f` applies the same UPSERT treatment:

| file | change |
|---|---|
| `SkillModel.js` | +40/−2 — UPSERT; preserves `created_at`, `source_plugin`, `is_user_modified`; `user_id` now insert-only |
| `CustomToolModel.js` | +32/−2 — UPSERT; preserves `created_at`; `created_by` now insert-only |
| `SkillModel.provenance.test.js` | +279 — 10 tests, 5 fail on the old statement |
| `CustomToolModel.provenance.test.js` | +197 — 6 tests, 2 fail on the old statement |

**The skills one is the reason this PR grew.** It isn't lost metadata like the agents case — it's silent data loss. `SkillService.updateSkill` sets the PRD-057 flag *before* it saves, so the save erased the flag it had just been told to set:

```js
if (existing.source_plugin) UPDATE skills SET is_user_modified = 1 ...
await SkillModel.createOrUpdate(id, merged, userId);   // ← wipes it back to 0
```

`PluginAssetLoader._decideUpdate` reads that flag to choose skip-vs-overwrite. Pinned at `0`, with `source_plugin` nulled a moment later, **the next upgrade of a plugin overwrote the user's edited skill without a word.** Neither guard could ever fire.

The tools case is the mild one — `tools` has no provenance flags, so an edit only reset `created_at`.

---

### Two things worth a second opinion

**1. `user_id` is now insert-only on `skills` — a small behaviour change.**

All three statements rewrote the owner column from the caller's argument. On agents and tools that was unreachable (both services fork to a new id when the editor isn't the owner). On skills it *was* reachable: `SkillService.updateSkill` looks the row up with an unscoped `findById` and passes the **requester's** id, so editing someone else's skill transferred ownership to them.

Making it insert-only stops the transfer. It does **not** fix the underlying gap — a non-owner can still edit the contents. That's a `SkillService` change, and I deliberately left it out rather than half-fix an authorization bug inside a persistence PR. Happy to do it as a follow-up if you'd rather it were one change.

**2. `is_builtin` is in the skills `SET` list.**

I kept it updatable because callers legitimately pass it. If a built-in skill should never flip to user-authored via a save, it belongs in the insert-only set instead — that's a product call, not one I should make unilaterally.

---

### Callers I checked

`SkillModel.createOrUpdate`: `SkillService` ×3 (`:56`, `:230`, `:275`), `SkillEvolver` ×3 (`:75`, `:158`, `:323`), `SkillDiscoveryRoutes:132`.
`CustomToolModel.createOrUpdate`: `ToolService:37`.

`SkillEvolver:323` is the interesting one — it round-trips a DB row (snake_case) back through the model, which accepts both casings. Unaffected by this change, but it's the call worth a glance.

---

### Verification

- **14 of the 30 new tests fail on the old statements** (7 agents + 5 skills + 2 tools), real SQLite in a throwaway `AGNT_HOME`. Mocked `db.run` can't catch this — the defect *is* the row-deletion semantics.
- **Mutation-tested 8/8 killed**, including reverting each statement and simulating a future migration to prove the ratchets fire.
- CI green on `16998e6f`: all 5 checks.
- Deployed to a live install and verified end-to-end: a row stamped with all five provenance columns survives a save through the running server unchanged, while the fields the caller supplied still update.

No schema change, no migration, no frontend. Six files, all under `backend/src/models/`.

agnt-gg:
The four commits were already applied to `main` by rebase (75b40a0a, 35d23b34) to keep history linear, so the original SHAs were not ancestors of `main` and GitHub could not auto-close this. Merging properly now so the PR is recorded as merged and you get the credit.

This merge is content-neutral — verified locally: merging both PRs onto the current tip yields tree `381ad14a`, identical to `main`'s tree, with an empty `git diff`. Patch-ids matched on all four commits.

Verified before landing: 4262 backend tests (52 new here, **0 regressions**), 221 frontend files, node runner byte-identical to baseline. The provenance fix was additionally proven against the real production schema — `created_at`, `source_plugin`, `is_user_modified`, `insight_version` and the owner column now all survive an edit.

Thanks @rimusz — tests that fail on the old code made this trivial to trust.

PR #71, opened 2026-08-22
fix(workflow): an unreachable workflow process is not a workflow state

rimusz:
Follow-up to #70, which named this as the reason that bug took so long to diagnose. Independent of it — test-merged both ways, `Auto-merging backend/src/services/WorkflowService.js / Automatic merge went well`.

## What

`WorkflowProcessBridge.fetchWorkflowState` answered **every** IPC failure with `{ status: 'error', error: error.message }`. That is wrong twice over, and this replaces it with a distinction the callers can act on.

### 1. `'error'` is a real workflow status

`ProcessWorker` sets a workflow's status to `'error'` when its engine fails (`ProcessWorker.js:33,61,109`), and `ProcessManager.fetchWorkflowState` reads that value back out of the database for an inactive workflow. So the bridge was answering *"I could not reach the workflow process"* with the identical value that means *"this workflow failed"*. No caller could tell them apart.

Both restart checks in `WorkflowService` test `['running','listening','queued'].includes(currentState.status)`. On an infrastructure failure the status is `'error'`, which is not in that list, so the branch is quietly skipped: **a live workflow silently keeps running its old definition after a save**, and the log line blamed a cause it had not checked.

The API is worse — `GET /workflows/:id/status` returned **200** with `{status:'error'}`, so a client saw a failed *workflow* rather than an unreachable *service*.

### 2. It discarded the child's own diagnosis

`sendMessage` rejects for two structurally different reasons:

| | cause | what we know |
|---|---|---|
| **transport** | never spawned, not ready, failed to initialise, no answer in time | nothing about the workflow |
| **reply** | the child answered `{ success: false, error }` | a specific, actionable diagnosis |

The second case carries messages like `Workflow wf-1 cannot be executed: node "n1" is missing "text"`. That text was replaced by the single word `error` before any caller saw it. **This is precisely why the #70 bug could not be diagnosed from the API** and had to be dug out of the workflow-process log.

### 3. The handler's own catch was unreachable

`WorkflowService.fetchWorkflowState` already had a catch testing `error.message.includes('not ready')` and returning `{status:'initializing'}`. It could never fire, because the bridge caught everything. Git dates the two: the swallow landed in `c9923d20` (2026-01-20), the handler in `2887115c` (2026-03-10). Someone wrote handling for an error that structurally could not arrive, and nothing said so.

## How

- **`sendMessage`** rejects transport failures with a typed `WorkflowProcessUnavailableError` carrying `reason` of `not-spawned` / `not-ready` / `init-failed` / `timeout`. A failure the child *reported* stays a plain `Error` with its message intact.
- **`fetchWorkflowState`** logs and rethrows — matching `restartActiveWorkflows` in the same class, which already does exactly this. All three callers already sit inside a `try/catch`, so control flow is unchanged wherever it was already correct.
- **`GET /workflows/:id/status`** now distinguishes:
  - `not-ready` → **200** `{status:'initializing'}` — unchanged intent, just finally reachable
  - other transport failures → **503** `{status:'unavailable', reason, details}`
  - child-reported failure → **500** with the diagnosis in `details` (house style, matching `saveWorkflow`)
- **Both restart-check catches** report the real error instead of asserting "not ready" for every cause — including a failure of the restart itself, which this block also swallows.
- `isWorkflowProcessUnavailable(error)` is exported so nobody has to sniff message text again. The old check matched one of the four transport failures and would have silently stopped matching on a reword.

## Testing

**22 tests across 2 files.** Without the change, **11 of the 13 bridge tests fail** and the **9 route tests cannot load at all** (the typed error does not exist). The bridge tests drive the real bridge with a mocked child, following `WorkflowProcessBridge.lifecycle.test.js`. The route tests deliberately do **not** mock the predicate or the error class — the wiring between them and the handler is part of what is under test.

The baseline reproduces the bug verbatim:

```
AssertionError: expected { status: 'error', …(1) } to be an instance of Error
AssertionError: expected { status: 'error', …(1) } to be undefined
```

**Mutation tested — 8 mutants, 8 killed, 0 survived:**

| mutant | result |
|---|---|
| restore the original swallow | killed (2) |
| untype the not-spawned reject | killed (2) |
| untype the timeout reject | killed (1) |
| untype the not-ready reject | killed (1) |
| predicate matches message text instead of code | killed (5) |
| drop the 200 "initializing" branch | killed (1) |
| stop forwarding the child's diagnosis | killed (1) |
| classify a child-reported failure as unavailable | killed (2) |

Full backend suite: **291 passed / 292 files**, including the existing `WorkflowProcessBridge.lifecycle.test.js` which exercises `sendMessage` directly. The single failure, `UpdateScheduler.status.test.js > replaces the previous pass rather than accumulating`, is pre-existing and unrelated — verified failing identically on untouched `origin/main` while investigating #70, passes in isolation, imports nothing this PR touches, and does not fire on CI.

## Notes for the reviewer

- **`/workflows/:id/status` can now return non-2xx.** Both frontend pollers already handle it: each does `if (!response.ok) throw`, and the surviving poller reschedules from its catch — the same thing it previously did on `data.status === 'error'`, which is in its continue-polling list. So client behaviour is unchanged in the failure case either way; what changes is that the failure is now *legible*.
- **Deliberately not included:** `activateWorkflow` and `deactivateWorkflow` swallow the same way, returning `{ error: error.message }`. That shape at least does not impersonate a status, so it is a milder version of this bug and a bigger blast radius. Happy to do it as a third PR.
- The two restart-check `catch` blocks still skip the restart when the process is unreachable. Retrying there is a behaviour change I did not want to fold into an error-reporting fix.

### DIFF
--- a/backend/src/services/WorkflowService.fetchWorkflowState.test.js
+++ b/backend/src/services/WorkflowService.fetchWorkflowState.test.js
@@ -1,0 +1,177 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * GET /workflows/:id/status must not report a workflow state it never read.
+ *
+ * The bridge answered an unreachable workflow process with
+ * `{ status: 'error' }`, and this handler passed that through as 200. Since
+ * 'error' is a genuine workflow status — ProcessWorker sets it on a workflow
+ * whose engine failed — a client had no way to tell "your workflow failed"
+ * from "I could not reach the workflow process".
+ *
+ * The handler's own `catch` was unreachable while that was true: it tested
+ * `error.message.includes('not ready')`, but nothing ever threw. It was added
+ * on 2026-03-10, two months after the swallow landed (2026-01-20) — someone
+ * wrote handling for an error that structurally could not arrive.
+ *
+ * The predicate and error class are deliberately NOT mocked; the wiring
+ * between them and this handler is part of what is under test.
+ */
+
+vi.mock('../services/auth/sessionTokenCache.js', () => ({
+  subscribe: vi.fn(() => () => {}),
+  getSessionToken: vi.fn(() => null),
+  getSessionUserId: vi.fn(() => null),
+}));
+vi.mock('child_process', () => ({ fork: vi.fn(), default: { fork: vi.fn() } }));
+
+const fetchWorkflowState = vi.fn();
+
+vi.mock('../workflow/WorkflowProcessBridge.js', async (importActual) => {
+  const actual = await importActual();
+  return {
+    ...actual,
+    default: {
+      fetchWorkflowState,
+      activateWorkflow: vi.fn(),
+      deactivateWorkflow: vi.fn(),
+    },
+  };
+});
+
+vi.mock('../models/WorkflowModel.js', () => ({ default: { findOne: vi.fn(), createOrUpdate: vi.fn() } }));
+vi.mock('../models/WebhookModel.js', () => ({ default: {} }));
+vi.mock('../models/database/index.js', () => ({
+  default: { run: vi.fn() },
+  dbRunWithRetry: vi.fn(async (fn) => fn()),
+}));
+vi.mock('../utils/realtimeSync.js', () => ({
+  broadcast: vi.fn(),
+  broadcastToUser: vi.fn(),
+  RealtimeEvents: { WORKFLOW_CREATED: 'created', WORKFLOW_UPDATED: 'updated' },
+}));
+vi.mock('../plugins/PluginManager.js', () => ({ default: {} }));
+vi.mock('../plugins/PluginInstaller.js', () => ({ default: {} }));
+
+const { default: WorkflowService } = await import('./WorkflowService.js');
+const { WorkflowProcessUnavailableError } = await import('../workflow/WorkflowProcessBridge.js');
+
+function makeRes() {
+  return {
+    statusCode: 200,
+    body: undefined,
+    status(code) {
+      this.statusCode = code;
+      return this;
+    },
+    json(payload) {
+      this.body = payload;
+      return this;
+    },
+  };
+}
+
+const req = { params: { id: 'wf-1' }, user: { userId: 'user-1' } };
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  vi.spyOn(console, 'error').mockImplementation(() => {});
+});
+
+describe('GET /workflows/:id/status — the process answered', () => {
+  it('passes a real workflow state straight through', async () => {
+    fetchWorkflowState.mockResolvedValue({ status: 'listening', outputs: { n1: 1 }, errors: {} });
+
+    const res = makeRes();
+    await WorkflowService.fetchWorkflowState(req, res);
+
+    expect(res.statusCode).toBe(200);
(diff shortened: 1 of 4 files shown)


rimusz:
Two CI notes for whoever picks this up.

**The first Playwright run failed; it was a flake, not this change.** `tests/e2e/agents.spec.js › can navigate to agents and see the list` timed out after 5s waiting for `Test Agent 1`. Evidence it is unrelated: this branch touches four backend workflow files and nothing the agents page loads; the same spec passed on #70 twenty minutes earlier at the identical base commit (`2bffa7e2`); and the suite passed on the second run here with the workflow code unchanged. The `@llm/reasoningPredicates.js` "could not be resolved" line in that log is pre-existing noise — it appears in #70's *passing* run too.

> ~~It does look like a genuine latent problem, though: `frontend/src/store/app/aiProvider.js:40` imports it, and the only `reasoningPredicates.js` in the tree is under `backend/src/services/ai/descriptor/`. Not touching it here.~~
>
> **Correction — that was wrong, see the follow-up comment below.** The alias is defined in `frontend/build/aliases.js` and shared by both `vite.config.js` and `vitest.config.js`; the file is tracked in git; and the module *is* inlined into the bundle. The warning is cosmetic and there is nothing latent here.

**Copilot could not review this PR** — it posted "Copilot encountered an error and was unable to review this pull request." I do not have permission to re-request a review on this repo (`RequestReviewsByLogin` is refused for my account), so someone with write access would need to click re-request if you want its pass. Happy to address anything it raises afterwards.

rimusz:
Correcting myself: the `@llm/reasoningPredicates.js` note in my comment above was wrong. **There is no latent bug.** I have struck that sentence in the original so nobody reads it alone and goes looking.

I called it "a genuine latent problem" on the strength of two checks, and both were bad:

| what I claimed | what is actually true |
|---|---|
| the `@llm` alias is undefined | it is defined in `frontend/build/aliases.js:27` and **shared by both** `vite.config.js:51` and `vitest.config.js:28` — I grepped only `vite.config.js` and missed the extracted module, which that file's own comment points at |
| the target file is not in the repo | `git ls-files backend/src/services/ai/descriptor/reasoningPredicates.js` returns it; it is present in a clean clone |

I then "confirmed" it by grepping `frontend/dist` for `supportsZaiReasoningEffort` and finding nothing — which proves nothing, because a production build **minifies exported identifiers**. Re-running it against string literals, which survive minification:

```
"adaptive"          -> found in 4 bundle files
"deepseek-chat"     -> found in 4 bundle files
"deepseek-reasoner" -> found in 4 bundle files
```

The module is inlined and the alias resolves.

The real explanation for the warning: `@llm` points **outside the Vite root** (`frontend/` → `backend/src/services/ai/descriptor`), so Vite's dependency **pre-bundling scanner** reports it as unresolved while `resolve.alias` handles it correctly in the actual module graph. The build exits 0, the frontend vitest suite passes — including `aiProvider.sharedDescriptor.spec.js`, which imports `@llm` directly — and the warning shows up on passing runs, including #70's. `build/aliases.js` already documents the arrangement and why it costs nothing at packaging time.

So: cosmetic warning, working setup, nothing to fix. Apologies for the noise — the rest of that comment (the Playwright flake analysis and the Copilot re-request) still stands.

PR #80, opened 2026-08-25
fix(auth): pin trust -> shape -> read across message handlers

rimusz:
> **Draft: stacked on #78 and #79.** The contract asserts that every message receiver calls an approved trust predicate, and two of those predicates are introduced by those PRs — neither is on `main` yet, so this can only run green on top of both. The diff below therefore includes their commits. Once they land, a rebase drops them and this reduces to **one new file, 698 lines**. Please merge those two first.

Three defects in this codebase came from getting one sequence wrong inside a `message` receiver, and each was found only after shipping:

1. **No trust check at all.** `Connectors.vue` redeemed an OAuth authorization code — with the signed-in user's bearer token — from any sender whatsoever.
2. **A trust check that did not work.** Two handlers shared `allowedOrigins.some((origin) => event.origin === origin || event.origin.includes('localhost'))`, whose second term never references the callback's own parameter. That makes it an unconditional substring test, admitting `localhost.evil.com` and anything like it.
3. **Trust without shape.** Handlers that had cleared the origin then read `event.data.type` directly, so a trusted-origin sender posting `null` threw a `TypeError`. The handlers are `async`, so it surfaced as an unhandled rejection: nothing crashed, nothing was logged anywhere anyone looks, and the handler silently did not run.

Every one of those is an **ordering or presence property of a handler body**. None is visible in review, none turns a test red on its own, and two shipped for months. That is the shape of defect a mechanical check earns its keep against.

## Two kinds of receiver, and why the difference decides the rule

Discovery matches both shapes a receiver can take, and they are not treated alike:

| | | Trust rule |
| --- | --- | --- |
| **GLOBAL** | `window.addEventListener('message', h)`<br>`window.onmessage = h` | Anyone who can reach the window can post, so the handler body is the entire boundary → **must check** |
| **HANDLE** | `someChannel.onmessage = h` | A sender must already hold the object; possession is the boundary, settled at construction → **exempt** |

A handle receiver has no `event.origin` to compare and no `event.source` to identify, so demanding a check would be demanding one it cannot perform. The exemption is **derived from the syntax, never granted by name** — and note the second line above: `window.onmessage = h` is classified *global* precisely so a receiver cannot buy the exemption by changing shape.

Shape discipline applies to both. A malformed payload throws the same way whichever door it came through.

**This gap was real, not hypothetical.** #78 gained a BroadcastChannel fallback partway through review, and with it a `channel.onmessage =` receiver. The contract discovered only `addEventListener`, so that receiver sat in the tree unaudited while this suite stayed green — the precise drift this file exists to prevent, happening underneath it.

## The rule is trust → shape → read, not origin → shape → read

Two of the six global receivers never look at `event.origin`, and are right not to. They compare the **sender** instead:

```js
widgetSdk.js    widgetWindows.has(event.source)                    // a WeakSet allowlist
Artifacts.vue   event.source !== previewFrame.value.contentWindow  // reference identity
```

Identity is strictly stronger than origin. Same-origin does not mean *"the window I opened"* — which is precisely the hole Copilot found in the Google sign-in handler, since this app renders authored HTML in `allow-scripts allow-same-origin` iframes. Demanding a redundant origin comparison there would be cargo-cult, so **either mechanism satisfies the contract**.

Which receivers are origin-gated is **derived from the code, never listed**. Convert an identity-gated one to origin-gating and the ordering rule starts applying to it automatically, with nobody having to remember this file exists.

## A handle receiver is not left unguarded either

Its boundary is *which object it was handed*, and for a `BroadcastChannel` that is decided entirely by the name. Two sides typing the same string is the same copy-paste failure the shared origin predicates exist to prevent — except this one **fails silently**: a mismatched name is not an error, it is a channel nobody ever posts to.

So it gets the handle equivalent of *"leave origin comparison to the shared predicates"* — leave the channel name to the shared constant. Any `new BroadcastChannel('literal')` fails the suite.

## What it deliberately does not check

That a trust check is **correct**. A predicate can be present and still be wrong — defect 2 above was exactly that. Correctness is pinned by each predicate's own unit tests; this pins that one is *called*, and called *first*. The two jobs are complementary and neither substitutes for the other.

## Scope

**Seven receivers** are covered: six global and one handle. Two other `addEventListener('message')` registrations are deliberately not matched — `dc.addEventListener` in `useRealtimeVoice.js` is a negotiated WebRTC peer, and `eventSource.addEventListener` in `appAuth.js` is a same-origin SSE stream. Neither is third-party spoofable and neither carries an `event.origin`.

The widget-side listener at `widgetSdk.js:161` is excluded because it lives inside a template literal, is injected into a widget iframe's `srcdoc`, and runs over there rather than in this application. That exclusion works by **blanking template-literal contents, not by naming the file**, so it cannot rot — lift that code into the host and it comes back under every rule. It would pass anyway: it opens with `if (event.source !== parent) return;`. Three tests pin this.

## Validation

Rather than assert the contract works, I ran it against the real states of the tree:

| Tree state | Result |
| --- | --- |
| `main` today (neither PR) | 3 rules fail, naming **all four** unguarded handlers — and distinguishing which had *no approved check* from which *compared the origin by hand* |
| #78 only | names exactly the **three OAuth** handlers |
| #79 only | names exactly the **sign-in** handler |
| both (this branch) | **26 tests green** |

That is an independent rediscovery of the same defect distribution the manual audit found.

Mutation tested, each hitting only what it should:

| Mutation | Fails |
| --- | --- |
| comment stripper → identity | 2 — its self-test *and* the anti-vacuity discovery test |
| template blanker → identity | 2 — the widget exclusion *and* the index-preservation self-test |
| `.onmessage` discovery matches nothing | 2 — receivers drop 7 → 6, and the both-kinds test names the gap |
| channel misclassified as global | 3 — **including the trust rule firing on it**, proving the exemption is load-bearing rather than decorative |
| channel name inlined as a literal | 1 — the shared-constant rule, alone |

The first two mutations target the contract's **own machinery**: a normaliser that eats real code would turn every assertion into a false pass.

There is also an explicit anti-vacuity test — discovery must find ≥ 7 receivers, of both kinds, and name the expected files. Renaming `addEventListener` or breaking the body extractor therefore cannot empty the set and let everything pass silently.

Nothing in the tree currently writes `window.onmessage =`, so the global/handle classification would otherwise never be exercised by real code. It is covered directly, in the same spirit as the normaliser self-tests, rather than waiting for someone to write that line.

Full frontend suite on this branch: **4153 tests green**.

### DIFF
--- a/frontend/src/__guards__/messageHandlerContract.spec.js
+++ b/frontend/src/__guards__/messageHandlerContract.spec.js
@@ -256,17 +256,87 @@ function walk(dir, found = []) {
 }
 
 /**
- * Every `window.addEventListener('message', ...)` in host code, with the body
- * of the function it registers.
+ * HOW A SENDER REACHES A RECEIVER, AND WHY THAT DECIDES WHAT TO DEMAND.
+ *
+ * Two shapes exist in this codebase and they are not equivalent:
+ *
+ *   window.addEventListener('message', h)   A GLOBAL receiver. Anyone able to
+ *   window.onmessage = h                    reach the window can post to it —
+ *                                           a page that framed us, an
+ *                                           extension, an allow-same-origin
+ *                                           artifact iframe. Receiving a
+ *                                           message implies nothing whatever
+ *                                           about the sender, so the handler
+ *                                           body is the entire trust boundary.
+ *
+ *   someChannel.onmessage = h               A HANDLE receiver. To post, a
+ *                                           sender must already hold the same
+ *                                           object: a BroadcastChannel on our
+ *                                           origin, a socket we opened, a port
+ *                                           we were handed. Possession IS the
+ *                                           boundary, and it was settled when
+ *                                           the object was constructed — there
+ *                                           is no `event.origin` to compare and
+ *                                           no `event.source` to identify.
+ *
+ * So the sender-trust rule applies to global receivers only, and that
+ * exemption is DERIVED from the syntax rather than granted by name. Note the
+ * second line above: `window.onmessage = h` is a global receiver wearing
+ * handle clothing, and is classified global for exactly that reason — it must
+ * not be able to buy the exemption by changing shape.
+ *
+ * Shape discipline applies to both. A malformed payload throws the same way
+ * whichever door it came through.
  *
  * Registrations on a DataChannel or an EventSource are deliberately not
  * matched: `dc.addEventListener('message')` is a negotiated WebRTC peer and
  * `eventSource.addEventListener('message')` is a same-origin SSE stream, so
- * neither is spoofable by a third party and neither has an `event.origin` to
- * check. Only `window` receives postMessage.
+ * neither is third-party spoofable and neither carries an `event.origin`.
  */
-function messageHandlers() {
-  const handlers = [];
+const GLOBAL_TARGETS = new Set(['window', 'globalThis', 'self']);
+
+/** Which kind of receiver an `X.onmessage =` assignment creates. */
+function receiverKindFor(target) {
+  return GLOBAL_TARGETS.has(target) ? 'global' : 'handle';
+}
+
+/**
+ * The body of the function starting at `at`, written inline or by name.
+ *
+ * Returns null when neither shape is recognised, so a call site the extractor
+ * does not understand goes MISSING from the set rather than being counted as
+ * compliant — which the anti-vacuity test below is there to notice.
+ */
+function functionAt(code, at) {
+  const after = code.slice(at);
+
+  // Inline: `function (event) {`, `(event) => {`, `async event => {`
+  const inline = after.match(/^(?:async\s+)?(?:function\s*)?\(?[\w$]*\)?\s*(?:=>)?\s*\{/);
+  if (inline) {
+    const open = at + inline[0].length - 1;
+    const close = matchBrace(code, open);
+    return close > -1 ? { name: '<inline>', body: code.slice(open, close + 1) } : null;
+  }
+
+  // Named: resolve the identifier to its declaration. The character that
+  // follows differs by call shape — `, handler)` for a listener, `= handler;`
+  // for an assignment — so accept any of them.
+  const named = after.match(/^([A-Za-z_$][\w$]*)\s*[),;\n]/);
+  if (!named) return null;
+  const name = named[1];
+
+  const declRe = new RegExp(`(?:const|let|var|function)\\s+${name}\\b[\\s\\S]{0,200}?\\{`, 'g');
+  const decl = declRe.exec(code);
+  if (!decl) return null;
+
+  const open = decl.index + decl[0].length - 1;
+  const close = matchBrace(code, open);
+  return close > -1 ? { name, body: code.slice(open, close + 1) } : null;
+}
+
+/** Every message receiver in host code, tagged with how a sender reaches it. */
+function messageReceivers() {
+  const found = [];
 
   for (const full of walk(FRONTEND_SRC)) {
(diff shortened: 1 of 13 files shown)


rimusz:
## Merge order, verified rather than assumed

Short version: **merge #78 and #79 in either order, then un-draft this one.** Details below, since "should be fine" is not the same as "was tried."

### 1. Nothing is stale

`main` is at `2835b283`, and the merge-base of **both** #78 and #79 *is* that commit — 0 behind, so each is already rebased by definition:

```
#78  merge-base 2835b283 == main   behind: 0   ahead: 3
#79  merge-base 2835b283 == main   behind: 0   ahead: 2
```

I ran the rebase anyway, in throwaway worktrees, to confirm it rather than infer it:

```
#78  rebase → "HEAD is up to date."   a4df3d7f → a4df3d7f
#79  rebase → "HEAD is up to date."   9f7f9689 → 9f7f9689
```

Both no-ops, SHAs unchanged. **Deliberately not force-pushed** — with nothing to replay it would only discard the green CI runs, detach the review threads from their commits, and orphan this PR's stack, which references those exact SHAs.

### 2. They are clean against each other, not just against main

That was the part actually worth testing. Trial-merging both into `main` in sequence, in a scratch worktree:

| Step | Result |
| --- | --- |
| `main` + #78 | 5 files, +592/−8 — no conflict |
| … then + #79 | 7 files, +764/−9 — no conflict |
| conflict markers / unmerged paths | **0** |
| full frontend suite on the combined tree | **4114 passed** |

The last row is the one that matters: a clean textual merge does not mean two changes are semantically compatible. Here they are.

Order is irrelevant because the two PRs share **no files** — `comm -12` over their file lists is empty. That is also why this branch could stack on both with a plain merge commit and no conflict.

### 3. This PR's claim is verified against the tree that will actually exist

The body above asserts that #80 goes green once the other two land. Rather than leave that as a promise, I dropped the single new guard file onto the post-merge tree (`main` + #78 + #79) and ran it:

```
16 passed (16)
```

So the stack resolves as advertised.

### Recommended sequence

1. **#78** — `MERGEABLE`/`CLEAN`, 5/5 checks green
2. **#79** — `MERGEABLE`/`CLEAN`, 5/5 checks green, both Copilot threads resolved
3. **rebase this branch and un-draft** — the merge commits and their six commits drop out, leaving **one new file, 698 lines**

Steps 1 and 2 are interchangeable.

<sub>All scratch worktrees removed; no PR branch was modified in the course of checking this.</sub>

agnt-gg:
I checked this one for smuggled content before anything else, since a stacked PR carrying merge commits is a clean way to hide a hunk that exists in neither parent. It is clean — all twelve shared files are byte-for-byte their source branch, and the only thing unique to this PR is the one test file. No evil merge. Worth saying plainly rather than leaving implied.

The craftsmanship is real too: deriving the gated handlers from the source instead of a hardcoded list, self-testing the comment stripper, and deliberately *not* asserting that the predicates are correct. That last restraint is the right call and most people do not make it.

Closing it anyway, for a reason that is not about the code.

Adopting this makes an external contributor the author of a policy that every future frontend change in this repo has to satisfy: no new `message` listener without an approved predicate, no inline `event.origin` comparison anywhere, no bare `BroadcastChannel` literal. Those may well be rules we want. They are not rules we can take from outside.

The mechanical side reinforces it. It scans 771 files on every run to police 8 handlers, and six file paths are hardcoded into its anti-vacuity assertion — move any one of them in an unrelated refactor and an auth test goes red for a reason that has nothing to do with auth. That is the failure mode that teaches people to ignore a gate.

The idea is sound. If we want that class of guard, we will write it against our own layout.

— Annie

PR #22, opened 2026-01-31
feat: Token auth with Keychain support and auto-refresh

Seredeep:
## Summary

- Add macOS Keychain integration for secure credential storage
- Implement PKCE utilities for OAuth authorization flows
- Create TokenRefreshManager with exponential backoff (1s-16s)
- Add ClaudeCodeAuthManager with Keychain → JSON fallback chain
- Add CodexAuthManager with account hash computation
- Integrate token refresh into LlmService for claude-code provider
- Add AES-256-GCM encrypted storage for non-macOS platforms
- Add AuthLogger utility with token redaction
- Create AuthError class with recovery guidance
- Add AuthProfileStore for managing auth-profiles.json
- Include comprehensive unit tests for all components

**Implements:** US2 (Keychain Token Access), US3 (Auto Token Refresh)
**Partial:** US4 (Error Recovery - AuthError class only)

## Test plan

- [ ] Verify Keychain read for Claude Code credentials on macOS
- [ ] Verify Keychain read for Codex credentials on macOS
- [ ] Test JSON fallback when Keychain unavailable
- [ ] Confirm token refresh triggers before expiry (5min buffer)
- [ ] Validate exponential backoff on refresh failures
- [ ] Run unit tests: `cd backend && npm test`

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #81, opened 2026-08-25
fix(auth): find the Claude Code login the CLI moved to the keychain

rimusz:
## What

`ClaudeCodeAuthManager` now falls back to the macOS Keychain when `~/.claude/.credentials.json` yields nothing, so a Claude Code login made by a current CLI is visible to AGNT again.

Two files, +284/-7, no new dependencies.

## Why

Claude Code **2.1.231** stores its session in the macOS Keychain — generic-password service `Claude Code-credentials` — instead of writing `~/.claude/.credentials.json`. `ClaudeCodeAuthManager` only ever reads that file (`resolveClaudeCredentialsPath()`, line 23).

The result on an affected machine is a provider that looks broken and cannot be fixed from the UI:

```
$ claude auth status
{ "loggedIn": true, "authMethod": "claude.ai", "subscriptionType": "max" }

$ ls ~/.claude/.credentials.json
No such file or directory

GET /api/providers/claude-code/auth/status
→ { "available": false, "apiUsable": false }
```

Re-running `claude auth login` does not help, because the new CLI writes to the Keychain every time. The only workaround is to hand-write the credentials file.

## How

`readClaudeCredentialsFile()` is the single read choke point for all five callers (`getAccessTokenSync`, `getAccessToken`, `checkApiUsable`, `refreshAccessToken`, `logout`), so the fallback lives there and nothing else needed rewiring:

```js
} catch {
  // No readable file. A newer CLI may still hold the session in the keychain.
  return { credPath, data: readClaudeKeychainCredentials() };
}
```

Deliberate constraints:

- **The file stays authoritative.** The Keychain is read only when the file yields nothing.
- **Cannot make a working install worse.** Every failure path returns `null` — byte-for-byte what this function already returned.
- **Bounded.** `spawnSync` with a 2s timeout, so a wedged keychain daemon degrades to "disconnected" instead of hanging every token read.
- **Read-only.** Never writes to the Keychain, never logs a token value.
- **Opt-out.** `CLAUDE_CODE_DISABLE_KEYCHAIN=1` for anyone who would rather AGNT not shell out to `security`.
- **Only the `claudeAiOauth` block is adopted.** The same keychain item also carries an `apiKey` field; promoting that would quietly swap a subscription login for a billed console key.

### The disconnect regression this had to avoid

`logout()` deleted the credentials file. On a keychain-backed CLI a naive fallback would then answer with the session that was just disconnected — the provider pops back to connected and **disconnect becomes unreachable**.

`logout()` now leaves a disconnect marker in the file instead. That keeps the file readable, which keeps the fallback out of it; connecting again clears the marker. There is a test that fails if this is reverted.

## Testing

New colocated suite `ClaudeCodeAuthManager.keychain.test.js` — 10 tests, vitest, no network and no real keychain. This manager previously had no test file at all.

Covered: keychain used when the file is absent; file preferred over keychain; both absent → `null`; `apiKey`-only item ignored; OAuth block without an access token ignored; unparseable output ignored; opt-out env honoured; disconnect survives a live keychain session; disconnect with no file on disk; reconnect clears the marker.

```
backend/src/services/auth/  →  14 files, 219 tests passed
full backend suite          →  308 files, 4896 tests passed
```

Both halves of the change were mutation-tested to confirm the tests can actually fail:

| Mutation | Result |
| --- | --- |
| fallback reverted to `data: null` | 2 tests fail |
| `logout()` reverted to `fs.unlinkSync` | the disconnect guard fails |

**Testing note worth knowing:** the tests drive a stub binary through `CLAUDE_CODE_SECURITY_BIN` rather than redirecting `HOME`. Redirecting `HOME` also relocates the login keychain (`$HOME/Library/Keychains`), so every lookup fails with "item not found" — an artifact that looks exactly like the bug. That override is also what lets this suite run on the Linux CI runner.

## Notes for reviewers

- **Prior art:** #22 proposed Keychain support and was closed. It was a 3,708-line, 23-file change that also added a new token-refresh framework, an encrypted-storage layer and a second test runner (Jest). This PR is deliberately the opposite: the smallest change that fixes the reported symptom, in the existing test runner, with no new abstractions. If #22 was closed on the *idea* rather than the scope, say so and I will close this too.
- **`AntigravityAuthManager` says the keychain cannot be read** (`AntigravityAuthManager.js:30`, `config/oauthClients.js:47`). On macOS that is not the case for an item whose ACL was created by the CLI under the same user: verified prompt-free reads from a `launchd` GUI-domain agent, the same context the AGNT backend runs in. This PR does not touch Antigravity.
- **Honest limit:** verified on one machine (macOS 26.6.2, Claude Code 2.1.231, brew cask). I cannot promise every Mac reads without a TCC prompt, which is exactly why the lookup is non-blocking, time-bounded and silent on failure — worst case is today's behaviour.

### DIFF
--- a/backend/src/services/auth/ClaudeCodeAuthManager.js
+++ b/backend/src/services/auth/ClaudeCodeAuthManager.js
@@ -2,6 +2,7 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 import crypto from 'crypto';
+import { spawnSync } from 'child_process';
 import axios from 'axios';
 import generateUUID from '../../utils/generateUUID.js';
 import { getClientIdentity } from '../ai/clientVersions.js';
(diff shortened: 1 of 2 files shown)


rimusz:
Copilot's automated review errored out on this PR ("Copilot encountered an error and was unable to review this pull request"), so there's no review feedback to act on yet.

Could a maintainer re-request it? I tried from my side and got a 404 — I have fork access but not push access to this repo, and re-requesting a reviewer needs the latter.

All five checks are green in the meantime, and I'm happy to address whatever Copilot flags once it runs.

agnt-gg:
Closing this under the same policy as the others, but you did the work, so here are the two things I would want to know if it were mine.

First, the best decision in the PR: adopting only the OAuth block and never the `apiKey` from the same keychain item. That item can carry a billed console key, and silently swapping a subscription login for a metered one is exactly the kind of thing nobody notices until the invoice arrives. Right call, and you called it explicitly rather than by accident.

**Flaw 1 — the logout rewrite is unconditional, the keychain read is not.** The read is `darwin`-gated; the marker write runs on every platform. On Windows and Linux this delivers only the behaviour change and none of the benefit, and it *creates* `~/.claude/.credentials.json` where none existed, writing an AGNT-proprietary key into another tool’s config directory. Gating the marker write on `keychainLookupEnabled()` fixes it — `logout()` then keeps deleting the file exactly as it does today on the platforms where the fallback never runs.

**Flaw 2 — disconnect becomes a one-way door.** The marker is only cleared by `writeClaudeCredentials()`, which only your own connect flow calls. A keychain-backed CLI never touches that file. So: disconnect in the app, run `claude auth login`, and you are still reported disconnected — permanently, recoverable only through the UI or by hand-editing the file. That is the same symptom class the PR exists to fix, one step downstream. Clearing the marker when the keychain holds a session with an expiry *newer* than the disconnect would restore CLI-login recovery without making disconnect a no-op.

Both are small, and your existing test file is already shaped to catch them.

— Annie

PR #83, opened 2026-08-28
chore(browser): update browser-use to 0.13.8

MagMueller:
AGNT pins its Browser Use runtime for reproducible installs. This updates that pin from 0.13.7 to the current published release, 0.13.8, through AGNT's documented one-line upgrade path.

Checks:

- 75 focused Browser Use tests passed
- 5,119 full-suite tests passed; the two unrelated timing flakes passed on focused rerun (42/42)
- `normalize-eol` checked 2,546 attributed files with no rewrites
- `git diff --check` passed
- Fresh Python 3.13 install verified the runner imports, supported chat adapters, domain controls, runner compilation, and structured error output

### DIFF
--- a/backend/src/tools/library/actions/ai-browser-use.js
+++ b/backend/src/tools/library/actions/ai-browser-use.js
@@ -31,7 +31,7 @@ import { RUNNER_PY, RUNNER_VERSION, RESULT_SENTINEL } from './browserUseRunner.j
  * Bumping this constant is the whole upgrade procedure: the version check below
  * reinstalls when the venv disagrees.
  */
-export const BROWSER_USE_VERSION = '0.13.7';
+export const BROWSER_USE_VERSION = '0.13.8';
 
 /**
  * Packages the runner needs that browser-use does not already pull in.


MagMueller:
Nathan, one more thought: CLI 3.0 may fit AGNT’s chat-driven browser better than the nested Browser Agent. Your chat agent can drive the browser directly, removing the second agent loop. We haven’t run an AGNT-specific A/B yet, but I can wire a small comparison if useful.

agnt-gg:
Merged — thanks!

Verified the upgrade path rather than just the pin: AST-diffed both wheels across the 26 upstream API surfaces our runner depends on (26/26 intact), then ran the real `RUNNER_PY` against a venv upgraded **in place** from 0.13.7, since that's the path existing users take rather than a fresh install. Identical results before and after. Full suite matched your number exactly, and the `runJournal` failure is a pre-existing Windows timing flake — that file's last commit is literally "Make journal failure tests deterministic on Windows".

Worth calling out beyond currency: `fix(llm): read structured output from tool calls for Groq tool-calling models` (#5240) directly fixes our `output_model_schema` path, plus the aiohttp/pypdf security pins.

**On CLI 3.0 — you were right, and it shipped in this same push.** Some findings from wiring it up:

- It's already in the wheel we pip-install (`entry_points.txt` exposes `browser-use`/`bu`/`browser`), so no `uv tool install` and no new system dependency, despite what the docs suggest.
- `BU_CDP_WS`, not `BU_CDP_URL` — `get_ws_url()` returns the former verbatim, while the latter is resolved via HTTP `/json/version`. Our in-app browser bridge is a bare WebSocket serving no HTTP, so the documented variable blocks 30s then fails.
- The **default** daemon only. A named daemon (`BU_NAME`) gives itself a tab via `Target.createTarget`, which our single-webview surface refuses.
- Same reason the skill's "first navigation is `new_tab(url)`" guidance needed overriding to `goto_url` for that surface.

Scoped to chat, where the calling agent is already the loop — the nested agent there was pure overhead, and it now course-corrects between steps. The workflow node keeps the nested agent by design: node parameters are templated from trigger data (Discord, email, webhooks), and a node whose parameter is a program reopens the injection hole `browserUseRunner.js` was written to close. So it's an addition rather than a swap.

Appreciate the nudge on this one.

PR #84, opened 2026-09-02
fix(providers): a display-name fallback tier must still fire

rimusz:

A fallback tier set to **Cursor** or **Z.AI** is silently dropped from every provider chain that is ever built. It saves, it renders in the fallback editor as configured, and it never fires.

The fallback editors store a tier as the provider's *display name* (`FallbackProviders.vue`: `{ provider: <displayName> }`), but `isKnownProvider` validated against raw `PROVIDER_CAPABILITIES` *keys*. Eighteen of the twenty built-ins lowercase to exactly their key, so the mismatch is invisible. Two do not:

| Stored by the UI | Registry key | Result |
|---|---|---|
| `Cursor` | `cursor-cli` | dropped |
| `Z.AI` | `zai` | dropped |
| `Claude-Code`, `Grok-Build`, +15 others | match | fine |

Anyone who picks Cursor or Z.AI as a fallback has a failover tier that quietly does not exist. Nothing logs it — `buildProviderChain` does `if (!isKnownProvider(...)) continue;`.

This is the same bug class as `7525e2b9` ("Z-AI display name was lowercased to `z-ai` instead of resolved to `zai`"), which fixed it for model fetching on the frontend. This is the backend chain-builder instance of it.

A concrete four-tier chain, before and after:

```
CONFIGURED                                  BEFORE                 AFTER
Claude-Code / claude-opus-5                 claude-code   (t0)     claude-code   (t0)
Cursor      / cursor-grok-4.6-medium-fast   -- dropped --          cursor-cli    (t1)
Grok-Build  / grok-4.6                      grok-build    (t1)     grok-build    (t2)
<uuid>      / deepseek-v4-flash-0731        <uuid>        (t2)     <uuid>        (t3)
```

## How does it solve it?

`isKnownProvider` now defers to `getProviderConfig`, which already resolves display names via its `name`-based fuzzy match — and is what `createLlmAdapter` has used all along. The chain builder was simply stricter than the code it feeds.

Three details worth a reviewer's attention:

**The resolved key is gated on `PROVIDER_CAPABILITIES` membership** rather than returned raw. Admitting a provider the capability registry does not know would trade a tier that is silently *dropped* for one that is silently *dead* — a chain claiming depth it cannot deliver, which is worse. Today `getAllProviderKeys()` and `Object.keys(PROVIDER_CAPABILITIES)` are identical, so this gate is currently unobservable; `ProviderFallback.capabilityGate.test.mjs` mocks a divergence to prove it works, and carries a tripwire test asserting the two sets stay identical so the day they drift, a test says so.

**The chain stores the canonical key**, so the executor's `getTextModels(String(tier.provider).toLowerCase())` and `createLlmClient(tier.provider)` both resolve. Tier identity is compared canonically too, so a chain no longer fails `cursor` over to `Cursor` believing they are different providers.

**The model lookup deliberately keeps the original spelling.** This is the part I would most like checked. `resolveTierModel` snaps an unrecognised model to the provider's first *static* model. For a CLI provider that list is a stale floor, not an authority — the UI offers models fetched live from the CLI that the hardcoded list has never heard of. Passing the canonical key there would make it "recognise" `cursor-cli` and silently rewrite a live, user-chosen `cursor-grok-4.6-medium-fast` to a hardcoded `cursor-grok-4.5-high`, turning a fix for a dropped tier into a config rewrite nobody asked for. A tier with no model still gets a default, because the executor resolves one from the canonical key.

### Deliberately not done

That last point exposes an asymmetry I chose to leave alone: `Cursor` gets the trust-the-caller branch (its display name has no static list), while `Grok-Build` — whose display name *does* equal its key — hits the snapping branch and has its configured model rewritten. I hit this for real while rebasing: upstream shrank `grok-build`'s static list from `['grok-4.5','grok-4.6']` to `['grok-4.5']`, and a tier configured for `grok-4.6` now silently becomes `grok-4.5`. A control run on pristine `main` confirms that is pre-existing behaviour on the primary path, unrelated to this change, so widening its blast radius did not belong in this PR. `ProviderFallback.displayNames.test.mjs` documents the asymmetry explicitly rather than leaving it to be rediscovered. Happy to follow up separately if you want it addressed.

`Local` also still cannot be a fallback tier — the frontend offers it but it has no `PROVIDER_CAPABILITIES` entry at all. That is a missing capability entry rather than a name mismatch, so it is a different change; a test pins the current `null` so adding one trips.

Custom providers keep their UUID identity and pass through untouched.

## How did you verify it?

34 new tests across two files. Rather than only asserting they pass, I mutation-tested them — six mutations of the new code, each of which turns the suite red:

| # | Mutation | Tests failed |
|---|---|---|
| A | restore the original raw-key match | 2 |
| B | push the raw stored spelling instead of the canonical key | 6 |
| C | normalize the model lookup too (the overreach above) | 4 |
| D | compare tier identity textually instead of canonically | 3 |
| E | drop the capability gate | 2 |
| F | dedupe on the raw spelling | 1 |

Two of my tests initially survived their mutants and were worthless: D passed because the pair I chose (`cursor` vs `Cursor`) is one a naive comparison gets right by accident — it now runs `it.each` over four asymmetric spellings; and E passed because the gate is unobservable while the two key tables coincide — hence the mocked-divergence file. Both were rewritten until they bite.

```
$ npx vitest run backend/src/services/orchestrator/ProviderFallback
 Test Files  4 passed (4)
      Tests  97 passed (97)
```

**Full suites, and an honest note on them.** Both suites have failures on current `main` that this branch does not cause. I verified that by stashing these three files and re-running — identical counts:

```
BACKEND   npx vitest run
  with this change:  2 failed | 351 passed (353 files) — 4 failed | 5536 passed
  pristine main:     same 4 failures
  -> scripts/worktree.test.js (2), browserFallbackSurface.test.js (2)
     Neither imports ProviderFallback.

FRONTEND  npx vitest run --root frontend
  with this change:  9 failed | 238 passed (247 files) — 7 failed | 4265 passed
  pristine main:     9 failed | 238 passed (247 files) — 7 failed | 4265 passed
  -> byte-identical; this branch touches no frontend file.
```

Branched from `origin/main` at `acc865d1`, not from a stale local base. `node scripts/normalize-eol.mjs --check` reports 0 rewrites.

## Checklist

- [x] This does not modify auth, sessions, or credential handling
      (see [Restricted areas](../CONTRIBUTING.md#restricted-areas))
- [x] This is not a fix for a security vulnerability
      (those go to [private reporting](https://github.com/agnt-gg/agnt/security/advisories/new))
- [x] `npm test` passes — *no new failures; 4 pre-existing, control-verified above*
- [x] `npm --prefix frontend test` passes — *no new failures; 7 pre-existing, counts identical to pristine `main`*
- [x] New behaviour has tests, and I watched them fail before the fix
      (mutation table above — stronger than red-then-green, and it caught two of my own tests being worthless)
- [x] Docs updated if the surface changed — no user-facing surface change
- [x] Commit subjects are 72 characters or fewer (60)

### DIFF
--- a/backend/src/services/orchestrator/ProviderFallback.capabilityGate.test.mjs
+++ b/backend/src/services/orchestrator/ProviderFallback.capabilityGate.test.mjs
@@ -1,0 +1,77 @@
+import { describe, it, expect, vi } from 'vitest';
+
+/**
+ * `resolveProviderKey` gates its result on PROVIDER_CAPABILITIES membership
+ * rather than returning whatever `getProviderConfig` resolved.
+ *
+ * Today that gate is invisible: PROVIDER_CONFIGS and PROVIDER_CAPABILITIES hold
+ * exactly the same 20 keys, so anything the config table resolves is already in
+ * the capability table. Deleting the gate changes no observable behaviour and
+ * every other test still passes — which is precisely why it needs a test of its
+ * own rather than a claim in a comment.
+ *
+ * The gate exists for the day those two tables diverge. A provider present in
+ * the config table but absent from the capability table would otherwise be
+ * admitted to the chain and then die in the executor: a tier that is silently
+ * DEAD instead of silently DROPPED, which is worse, because the chain reports a
+ * depth it cannot actually deliver.
+ *
+ * This file mocks the config table to force that divergence. It lives apart
+ * from the main suite because the mock is module-wide.
+ */
+
+vi.mock('../ai/providerConfigs.js', async (importOriginal) => {
+  const actual = await importOriginal();
+  return {
+    ...actual,
+    getProviderConfig: (key) => {
+      // A provider the CONFIG table knows about but the CAPABILITY table does
+      // not — the exact divergence the gate is there to catch.
+      if (String(key).toLowerCase() === 'ghost-provider') return { key: 'ghost-provider', name: 'Ghost' };
+      return actual.getProviderConfig(key);
+    },
+  };
+});
+
+const PF = (await import('./ProviderFallback.js')).default;
+const ProviderRegistry = await import('../ai/ProviderRegistry.js');
+
+describe('resolveProviderKey — capability gate', () => {
+  it('the premise holds: the mocked provider is in the config table only', async () => {
+    const { getProviderConfig } = await import('../ai/providerConfigs.js');
+    expect(getProviderConfig('ghost-provider')).toMatchObject({ key: 'ghost-provider' });
+    expect(Object.keys(ProviderRegistry.PROVIDER_CAPABILITIES)).not.toContain('ghost-provider');
+  });
+
+  it('refuses a provider the capability registry does not know', () => {
+    expect(PF.resolveProviderKey('ghost-provider')).toBeNull();
+    expect(PF.isKnownProvider('ghost-provider')).toBe(false);
+  });
+
+  it('drops such a tier from the chain rather than shipping a dead one', () => {
+    const chain = PF.buildProviderChain({
+      provider: 'anthropic',
+      model: 'm',
+      fallbackEnabled: true,
+      fallbackProviders: [{ provider: 'ghost-provider', model: 'x' }],
+    });
+
+    expect(chain).toHaveLength(1);
+  });
+
+  it('still resolves real providers through the same path', () => {
+    expect(PF.resolveProviderKey('Cursor')).toBe('cursor-cli');
+  });
+});
+
+describe('config and capability tables — divergence tripwire', () => {
+  it('currently hold identical key sets', async () => {
+    // Not a requirement, an observation — but if it ever stops being true the
+    // gate above changes from dead code to load-bearing, and whoever caused the
+    // divergence should find out from a test rather than from a dead tier.
+    const { getAllProviderKeys } = await vi.importActual('../ai/providerConfigs.js');
+    expect([...getAllProviderKeys()].sort()).toEqual(
+      [...Object.keys(ProviderRegistry.PROVIDER_CAPABILITIES)].sort()
+    );
+  });
+});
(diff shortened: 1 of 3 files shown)


rimusz:
**Backend (vitest) is red, and it is red on `main` too — same two tests, same commit.**

Flagging it here so nobody has to dig, and because a red check on a PR that claims to be well-verified deserves an explicit account rather than a footnote.

This branch is based on `acc865d1`. That commit's own run on `main` ([33596904066](https://github.com/agnt-gg/agnt/actions/runs/33596904066)) fails the identical two tests:

```
main @ acc865d1        FAIL scripts/worktree.test.js > remove > detaches links first, …
                       FAIL scripts/worktree.test.js > remove > keeps an unmerged branch …
                       Test Files  1 failed | 350 passed (351)

this PR @ ff910668     FAIL scripts/worktree.test.js > remove > detaches links first, …
                       FAIL scripts/worktree.test.js > remove > keeps an unmerged branch …
                       Test Files  1 failed | 352 passed (353)
                             Tests  2 failed | 5538 passed (5540)
```

Same two failures either side; the delta is `+2` files and `+2` passing files, which are the two test files this PR adds. The underlying error is environmental rather than assertive — `git worktree remove` refusing on `alpha has uncommitted changes` — and `scripts/worktree.test.js` does not import `ProviderFallback`, `buildProviderChain` or `resolveProviderKey`.

Two corrections to my own PR body, now that CI has more authority than my laptop did:

- **Frontend passes cleanly in CI.** I reported 7 pre-existing frontend failures reproducing identically on pristine `main`. That was true locally but it was my environment — I ran the suite from a git worktree with a symlinked `node_modules`. CI is the trustworthy signal here, and it is green. Disregard that half of the note.
- **`browserFallbackSurface.test.js` also passes in CI**, for the same reason. The only real failure is `scripts/worktree.test.js`, and it predates this branch.

Happy to rebase onto a green `main` whenever that test is sorted, if you would rather see this PR fully green before merging.

agnt-gg:
Landed on `main` as 8f8455e5 (rebased onto 46679540, fast-forward, full backend suite 353 files / 5544 passed). The rebase changed the SHA, so GitHub cannot mark this merged automatically — closing by hand. Thank you, @rimusz.

Issue #88, opened 2026-09-03
chat: interrupted-tool-call settlement misses the mid-announce abort window


rimusz:
Found while porting cf9be1d12f into a fork. That commit settles open tool calls on abort so their cards stop spinning — it works for the window it describes, and there is a second window where the guard is false and nothing is emitted.

All line numbers below are pinned to 13d3dbce.

## Two abort windows, one covered

| abort lands | `toolCalls` | emission at 3522 |
|---|---|---|
| between rounds — a stream completed, next round pending | assembled | fires ✓ |
| mid-announce — calls still streaming in | empty | **guard false, nothing emitted** |

## Why

The emission is guarded on `toolCalls && toolCalls.length > 0`, but `toolCalls` is only assigned when a stream *completes*:

```
2526:          toolCalls = retryResponse.toolCalls;
3455:      toolCalls = nextResponse.toolCalls;
```

The event that actually draws the card fires much earlier, from the `tool_call_delta` handler as the model emits each call:

```
2238, 2512, 3444:  sendEvent('tool_pending', …)
```

`tool_pending` and `tool_start` both map to `CHAT_TOOL_START` client-side, so either one puts a spinner on screen. Abort while calls are still being announced and the cards exist while `toolCalls` does not, so the loop at 3522 is never entered.

## Reproduction

Ask for **several tool calls in one round**, then cancel while they are still being announced:

```
Call execute_shell_command FOUR TIMES IN PARALLEL, in a single turn,
all four at once: `sleep 6 && echo one`, `sleep 6 && echo two`,
`sleep 6 && echo three`, `sleep 6 && echo four`.
```

POST `/api/orchestrator/runs/:conversationId/cancel` once three `tool_pending` events have arrived, then drain the SSE stream to the end.

The shape of the reproduction is the non-obvious part, and two other shapes do **not** work:

- cancelling *during* a tool's execution — the call finishes and gets its `tool_end`;
- cancelling cleanly *between* rounds — the run stops before the model issues anything, so nothing is left drawn.

Both were tried first and looked like "no bug".

## Measured

Against a backend carrying cf9be1d12f (the block is byte-identical to yours):

```
cards drawn : 3      tool_end : 0      OPEN (spinners): 3
```

`[Tool Loop] Aborted with N open tool call(s)` is absent from the log, so the block was never reached rather than reached and ineffective. The stream also produced **zero** `tool_start` events, independently confirming those calls never entered the loop.

## Impact: nothing a user sees

[...]

PR #89, opened 2026-09-03
fix(chat): fail over mid-turn when the primary stream is overloaded

rimusz:

A primary provider that returns HTTP 529 (`overloaded_error`) after tools have already run ends the turn with an API error card. The configured fallback chain never gets a vote, because only the first stream of the turn was wrapped in `runWithFallback`.

## How does it solve it?

Every streaming LLM call in a chat turn (round 0, validation retry, tool loop, empty-response follow-up) now goes through `streamAcrossChain`, the same helper that rebuilds client+adapter per tier. Autonomous / goal messages use the same wrap on later rounds and the forced final reply.

Deliberately unchanged: adapter-level retries still run first; failover still only happens when the adapter reports `recoveredFromError` (zero tokens emitted on that tier). Settings are not persisted.

## How did you verify it?

```
npx vitest run   backend/src/services/orchestrator/toolLoopFailover.test.js   backend/src/services/orchestrator/primaryTierFailover.test.js   backend/src/services/orchestrator/ProviderFallback.test.mjs

Test Files  3 passed (3)
     Tests  63 passed (63)
```

Wiring tests assert `adapter.callStream` is only invoked from `runTierStream`. A simulated mid-turn 529 on the primary is served by the next tier.

## Checklist

- [x] This does not modify auth, sessions, or credential handling
      (see [Restricted areas](../CONTRIBUTING.md#restricted-areas))
- [x] This is not a fix for a security vulnerability
      (those go to [private reporting](https://github.com/agnt-gg/agnt/security/advisories/new))
- [ ] `npm test` passes
- [ ] `npm --prefix frontend test` passes
- [x] New behaviour has tests, and I watched them fail before the fix
- [ ] Docs updated if the surface changed
- [x] Commit subjects are 72 characters or fewer

Made with [Cursor](https://cursor.com)

### DIFF
--- a/backend/src/services/AutonomousMessageService.js
+++ b/backend/src/services/AutonomousMessageService.js
@@ -310,42 +310,27 @@ Be empathetic and suggest potential solutions or next steps if appropriate.`,
         loopMessages = sanitizeEmptyAssistantMessages(loopMessages);
 
         let roundResponse, rawToolCalls;
-        if (round === 0 && autoProviderChain.length > 1) {
-          // Round-0 failover: if the primary tier exhausts retries, roll over to
-          // the next tier and adopt its adapter for all subsequent rounds.
-          const { result: r0 } = await runWithFallback({
-            chain: autoProviderChain,
-            onFallback: ({ from, to, reason }) => {
-              log(`[AutonomousMessage] Provider failover: ${from.provider}/${from.model || '?'} → ${to.provider}/${to.model || '?'} (${reason})`);
-            },
-            runOne: async (tier) => {
-              if (!tier.primary) {
-                const np = String(tier.provider).toLowerCase();
-                // Resolve the tier's model; null means "provider default" — never
-                // carry the primary provider's model onto a different provider.
-                const { getTextModels } = await import('./ai/ProviderRegistry.js');
-                const tierModel = tier.model || getTextModels(np)?.[0] || tier.model;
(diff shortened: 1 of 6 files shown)


agnt-gg:
Landed on `main` as 354e3d86 + 0eef4b1b — your two commits, your authorship, rebased onto today's head. Thank you; one audited call site for the whole turn is the right shape, and the wiring test that counts it is what will keep it that way.

Closing rather than merging because `main` moved under the PR three times today on the same seam (2ff650c9, 292712b9, ac44e739) and the fix was needed before you'd be back online. The red badge is a mechanical consequence of the rebase, not a verdict on the work.

One line was carried across in the rebase: `streamAcrossChain` now passes `shouldStop: () => streamAbortController.signal.aborted` into `runWithFallback`. That guard landed earlier today in ac44e739 — the Anthropic adapter, on seeing the abort signal, breaks out of its event loop cleanly and *returns* `{ recoveredFromError: true }` rather than throwing `AbortError`, so a user pressing Stop was indistinguishable from an overloaded provider and walked the entire chain (+28s and two false `provider_fallback` events before the interrupted cards settled). Your consolidation is what makes that guard cover every stream in the turn instead of round 0 only — strictly better than what `main` had before. The `primaryTierFailover` pin was re-pointed to assert the predicate lives in the helper.

Verified live against the landed build: a mid-turn 529 rolls to the next tier, and Stop after that does not.

Also closed #88 via 2ff650c9, using the id→name ledger shape you sketched there.

PR #97, opened 2026-09-07
feat(themes): add Everforest, a theme that follows the desktop

tryingET:
## What

Adds **Everforest**, the first theme that follows the desktop's colour scheme
instead of pinning one face, and de-duplicates the theme-class logic it had to
touch to get there.

## Why

Every existing theme is one fixed palette. A terminal-shaped app tends to sit
next to a terminal that already flips at sunrise and sunset, and the app
staying dark at noon is the thing you notice. Everforest ships both faces, so
`everforest` is one entry in the picker that renders light by day and dark by
night.

## How

`everforest` is a single theme id with two faces. The store adds `.dark`
alongside `.everforest` when `prefers-color-scheme` is dark, and removes it on
the flip.

Three pieces of duplication were in the way, and each is now single-sourced:

| Before | After |
| --- | --- |
| `SET_THEME` and `initTheme` each carried their own copy of the body-class block | `applyThemeClasses(theme)` |
| `isDarkMode` derived inline from `!['light','rose'].includes(theme)`, in three places | `isDarkFace(theme)` |
| `Docs.vue` re-derived light/dark from the theme name to choose image variants | reads the `.dark` class the store already maintains |

That last one is why the de-duplication is not cosmetic: `Docs.vue` reading the
theme *name* cannot ever be right for a theme whose face depends on the
desktop. Reading the class the store owns makes it right for every theme,
including the ones added after this.

A `matchMedia('(prefers-color-scheme: dark)')` listener re-applies the classes
**and** the `isDarkMode` flag on a day/night flip. The listener is registered
once and is a no-op unless the active theme follows the system.

## Following the desktop is not enough on its own

A desktop with no day/night switch answers `prefers-color-scheme: light`
forever. If following the system were the only mechanism, those users would
reach Everforest's light face and nothing else — no control in the UI would get
them to the dark one. So the face is a preference, not just an observation:

`themeFace` is `'auto' | 'light' | 'dark'`, persisted alongside the theme.
`'auto'` asks the desktop. The other two outrank it, and the watcher stands
down while one is set.

The control is the dark-mode toggle **already** in Settings. It used to flip
the `.dark` class directly, which for a system-following theme the next
re-apply silently undid; it now pins the face. Re-picking the theme in the
picker resets to `'auto'` — the way back, without a third control in the UI.

Single-face themes keep their existing behaviour on both paths, and
`theme.systemFace.spec.js` covers that so it stays true.

## The light ink is not Everforest's own foreground

Everforest Light Medium's foreground is `#5c6a72`. This codebase dims body copy
with `color: var(--color-text); opacity: 0.78 … 0.85`, and
`secondaryInkContrast.spec.js` composites that over both the page fill and the
raised card:

| ink | worst pairing |
| --- | --- |
| `#5c6a72` (Everforest Light fg) | 3.11:1 |
| `#4f585e` | 3.75:1 |
| **`#374145`** (shipped) | **4.90:1** |

So the light face uses the darker slate. With Everforest's own foreground the
gate failed across the marketplace, schema, plan and stat surfaces; with
`#374145` it is green.

## Tests

`npm --prefix frontend test -- --run src/styles src/store/app/theme.*.spec.js`
→ **13 files, 142 tests, all passing**, including `themeCanvas`,
`secondaryInkContrast`, `themeTokens`, `themeSurfaces`, `onFillContrast` and
the new `theme.systemFace.spec.js` (7 tests, with an anti-vacuity assertion
that the desktop watcher is actually registered before the flip is fired).

The full frontend suite reports 93 failed files / 848 failed tests both **with
and without** this branch — identical counts, measured against `origin/main` in
a clean worktree. They come from `user.config.js` reading `localStorage` at
module scope under the node environment, and are untouched by this change.

## Known boundary, worth a look on your side

`secondaryInkContrast.spec.js` resolves a theme's tokens with "last declaration
in the file wins". A one-file two-face theme therefore only ever gets its
*second* block measured — here, the light face. The dark face is unmeasured by
that guard. I checked it by hand against the same maths: worst pairing 5.02:1,
comfortably over AA. But the guard has a blind spot for this theme shape, and
it is yours to decide how to close.

### DIFF
--- a/frontend/src/store/app/theme.js
+++ b/frontend/src/store/app/theme.js
@@ -6,29 +6,40 @@ const SUPPORTED_THEMES = ['light', 'dark', 'cyberpunk', 'midnight', 'ember', 'no
 /* Themes that carry their own light and dark face and follow the desktop's colour scheme. */
 const SYSTEM_FOLLOWING_THEMES = ['everforest'];
 
+/* Which face of a system-following theme to show. 'auto' asks the desktop; the other two are an
+   explicit choice by the user, which is the only thing that works on a desktop that has no
+   day/night switch to ask -- there, `prefers-color-scheme` answers light forever. */
+export const THEME_FACES = ['auto', 'light', 'dark'];
+
+function readStoredFace() {
+  const stored = localStorage.getItem('themeFace');
+  return THEME_FACES.includes(stored) ? stored : 'auto';
+}
+
 function prefersDarkScheme() {
   return typeof window !== 'undefined' && typeof window.matchMedia === 'function'
     ? window.matchMedia('(prefers-color-scheme: dark)').matches
     : true;
 }
 
-/* The legacy isDarkMode flag. A system-following theme is dark only while the desktop is. */
-function isDarkFace(theme) {
-  return SYSTEM_FOLLOWING_THEMES.includes(theme)
-    ? prefersDarkScheme()
-    : !['light', 'rose'].includes(theme);
+/* The legacy isDarkMode flag, and the one rule for which face is showing. */
+function isDarkFace(theme, face = 'auto') {
+  if (!SYSTEM_FOLLOWING_THEMES.includes(theme)) return !['light', 'rose'].includes(theme);
+  if (face === 'dark') return true;
+  if (face === 'light') return false;
+  return prefersDarkScheme();
 }
 
 /* One place that decides which classes a theme puts on <body>, so startup and later switches
-   cannot drift apart. A system-following theme adds `.dark` only while the desktop asks for it. */
-function applyThemeClasses(theme) {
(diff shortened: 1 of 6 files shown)


agnt-gg:
Thanks, @tryingET. Nathan has visually reviewed and approved Everforest, and I’m bringing in your two original commits unchanged.

The six changed files match the approved preview exactly. The current-main theme/style checks pass all 138 tests; the preview also passed the full frontend suite and production build. I’m now waiting for both this PR’s checks and a separate current-main integration run before merging.

I’ll use GitHub rebase-and-merge on this original PR so it receives the purple Merged status and your authorship remains credited. #99 remains a separate follow-up.

— Annie

PR #98, opened 2026-09-07
test(themes): guard every rgb triplet, not just the canvas one

tryingET:
## What

Extends the drift guard that already exists for `--color-background-rgb` to
**every** hand-maintained rgb triplet in the theme files. Test only — no theme
file changes, no behaviour change.

## Why

Each theme writes its colours twice:

```css
--color-green: #22c55e;   /* for color:, border:, background: */
--green-rgb: 34, 197, 94; /* so a rule can write rgba(var(--green-rgb), 0.1) */
```

CSS cannot take an alpha of a hex token, so the triplet is decomposed by hand,
in the file, next to the hex. Two spellings of one value drift: change the hex,
forget the triplet, and every *solid* use of the colour moves while every
*translucent* one stays behind. The page still renders, so nothing notices.

`themeCanvas.spec.js` already pins exactly this invariant — for one token.
Measured across the theme files on `main`:

| | pairs |
| --- | --- |
| rgb/hex pairs declared | **42** |
| guarded by `themeCanvas.spec.js` | 7 (`--color-background-rgb`) |
| **unguarded** | **35** |

And the unguarded ones are not obscure:

| token | files consuming it |
| --- | --- |
| `--green-rgb` | 132 |
| `--primary-rgb` | 74 |
| `--red-rgb` | 27 |
| `--blue-rgb` | 19 |
| `--yellow-rgb` | 15 |

**No triplet is currently drifted.** This is a guard, not a repair — it holds
the invariant that happens to be true today.

## One detail worth a look

The collector reads each theme file **block by block** instead of running a
regex over the whole file. `themeCanvas.spec.js` takes the first match and
`secondaryInkContrast.spec.js` takes the last; either is fine while every theme
is one selector block, and both quietly measure half the file the moment a
theme carries two — a light face and a dark face in one file, say. Reading
blocks removes that assumption from this guard.

## Tests

`npm --prefix frontend test -- --run src/styles/themeRgbPairs.spec.js` → 5
passing, including two "has teeth" cases: a drifted triplet is detected, and a
second selector block is not skipped.

## Possible follow-up, your call

The duplication itself could be removed rather than guarded. Chromium supports
relative colour syntax, so `rgba(var(--green-rgb), 0.1)` could become
`rgb(from var(--color-green) r g b / 0.1)` and the `*-rgb` tokens could be
deleted outright. That is **1610** `rgba(var(--*-rgb), …)` call sites and an
architectural decision, so I have not assumed it — but if you want it, this guard is the safety net for
doing it incrementally.

### DIFF
--- a/frontend/src/styles/themeRgbPairs.spec.js
+++ b/frontend/src/styles/themeRgbPairs.spec.js
@@ -1,22 +1,23 @@
 /**
- * Guard: every hand-maintained rgb triplet still matches the hex it mirrors.
+ * Guard AGNT's base palette and theme RGB triplets against their matching colours.
  *
  * WHY
  * ───
- * The themes carry two spellings of the same colour. `--color-green: #22c55e`
- * for `color:` and `border:`, and `--green-rgb: 34, 197, 94` so a rule can
- * write `rgba(var(--green-rgb), 0.1)` — CSS cannot take an alpha of a hex
- * token, so the triplet is decomposed by hand, in the file, next to the hex.
+ * AGNT carries two spellings of the same colour: `--color-green: #19ef83`
+ * for solid use and `--green-rgb: 25, 239, 131` for existing
+ * `rgba(var(--green-rgb), 0.1)` consumers. This guard checks that contract;
+ * it does not change the palette or migrate consumers to newer CSS syntax.
  *
  * Two spellings of one value is a drift generator: change the hex, forget the
  * triplet, and every solid use of the colour moves while every translucent one
  * stays behind. Nothing in the app can notice — the page still renders.
  *
- * `themeCanvas.spec.js` already pins exactly this invariant, but only for
- * `--color-background-rgb`. Measured across the theme files at the time of
- * writing: 50 rgb/hex pairs, of which 7 were guarded. The other 43 were not,
- * and they are not obscure — `--green-rgb` is consumed in 132 files and
- * `--primary-rgb` in 74. This closes that gap for all of them.
+ * `themeCanvas.spec.js` already guards literal canvas pairs. Accents also
+ * need protection, including root brand colours and light/dark aliases.
+ * The bounded resolver below follows plain var(--token) chains in palette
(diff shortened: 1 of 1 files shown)


tryingET:
I investigated the relative-color follow-up before implementing it. My statement that no triplet was drifted was too broad: this guard checks literal pairs declared together, not inherited/runtime equality. Several themes override orange/violet colors while inheriting the base triplets. That is a visual difference to discuss, not an assumed fix.

There are also compatibility boundaries:
- Custom-background mode makes `--color-background` transparent while retaining the solid triplet for panels and screenshots.
- Saved/generated widgets consume the exported RGB tokens.
- In an isolated Chromium fixture using the bundled html2canvas, legacy rgba capture succeeds, but relative RGB computes to `color(srgb …)` and capture throws. `color-mix()` has the same failure. This is library-level reproduction, not full-app export testing.

I preserved the findings, source-hashed observations and runnable probes in a **research-only draft on my fork**: https://github.com/tryingET/agnt/pull/1. It makes no production changes and is not an upstream migration request.

I suggest keeping the compatibility tokens for now. Would tighter coverage of this guard's boundaries be useful, or would you prefer leaving this at the current test-only scope?

agnt-gg:
Thanks, @tryingET — this is a useful preventive guard, and I’m keeping your original commit and attribution intact.

I added one small maintainer commit in the same test file to cover the root brand palette, default light/dark aliases, malformed values, and negative cases through the actual collector. No production colours, dependencies, or consumer code changed. The coverage boundary is explicit: custom themes are checked for their own RGB declarations, rather than claiming a complete CSS-cascade audit.

The corrected suite passes all 16 tests; the current-main integration candidate also passes the full frontend suite (4,428 tests) and production build. GitHub’s remaining checks are running. Once those pass, I’ll use GitHub’s rebase-and-merge so this PR is recorded as merged and your contribution stays credited.

Appreciate the contribution!

— Annie

PR #99, opened 2026-09-07
fix(themes): one system decides what a theme looks like

tryingET:
> **Stacked on #97.** Until that merges, the diff here also shows its commits.
> The change in *this* PR is the last commit, `fix(themes): one system decides
> what a theme looks like`.

## What

There is one theme, and there were two systems painting it. This collapses them
into one and fixes the two bugs the split was causing.

## The two systems

| | decides by | persists as | owns body classes |
| --- | --- | --- | --- |
| **A** | `currentTheme` → `SET_THEME` → `applyThemeClasses` | `currentTheme` | yes |
| **B** | `isDarkMode` / `isCyberpunkMode` → `SET_DARK_MODE` / `SET_CYBERPUNK_MODE` | `darkMode`, `cyberpunkMode` | **also yes** |

B never touched `currentTheme`, and nothing ever read `darkMode` or
`cyberpunkMode` back — state is re-derived from `currentTheme` on every load. So
the two Settings toggles wrote into the system the other one overwrote.

## The two bugs that produced

**1. The cyberpunk toggle does nothing from a light theme.** Every dark variant
is `body.dark.<name>` in CSS:

```
_cyberpunk.css   body.dark.cyberpunk
_nord.css        body.dark.nord
_midnight.css    body.dark.midnight
_ember.css       body.dark.ember
_hacker.css      body.dark.hacker
```

`SET_CYBERPUNK_MODE` added `.cyberpunk` **alone**. From `light` or `rose` there
is no `.dark`, so the selector matched nothing and the click was invisible.

**2. Either toggle is reverted on reload.** `initTheme` re-applies classes from
`currentTheme`, which the toggles never updated.

## The fix

Both toggles choose a **theme** — the thing that persists and the thing the
classes are derived from. `SET_DARK_MODE` stays as the mirror components read,
without owning classes or storage.

`toggleDarkMode` keeps the behaviour #97 gave it for a theme carrying both
faces: that one switches *face*, not theme.

## Removed

`SET_CYBERPUNK_MODE`, `initDarkMode`, `initCyberpunkMode` — all unreachable
after the above, and none referenced anywhere in the repo. Happy to restore
them as no-op shims if you'd rather keep the surface.

## Tests

`theme.oneSystem.spec.js`, 6 cases: cyberpunk reachable from a light theme with
**both** classes; that it survives a simulated reload; the dark toggle moving
the theme so the name can't contradict the class; no write-only keys; the
two-faced theme still switching face; and the removed surface staying gone.

Full run: `src/styles` + the three `theme.*.spec.js` → **15 files, 153 tests,
all passing**.


agnt-gg:
Thanks, @tryingET. I preserved your original commit and its single-theme-state correction. Our current Settings screen no longer uses the legacy DarkModeToggle component, so I added a small maintainer follow-up removing that component and its unused actions/Cyberpunk mirror rather than repairing unreachable controls.

The live dark-mode mirror, current picker, Everforest system switching, backgrounds, and greyscale remain intact. Regression tests now exercise the current selection/persistence path and verify the obsolete entry points are absent. This is cleanup only; no new theme UI or palette changes.

The reviewed local candidate passed 209 focused checks, 4,472 frontend tests, and the production build. The updated original PR is running GitHub's required checks before rebase-and-merge, preserving your credit and merged status.

— Annie

PR #100, opened 2026-09-07
feat(providers): add Atlas Cloud template

binyangzhu000-sudo:
## What problem this solves
AGNT supports custom OpenAI-compatible providers, but Atlas Cloud users currently need to enter the endpoint and capabilities manually. This adds an opt-in template that pre-fills the compatible endpoint and a current multimodal default model.

The change stays within the generic provider-template path and does not modify authentication, credential storage, or other restricted areas.

## Verification
- `npx --yes vitest@4.0.18 run backend/tests/providers/providerConfigs.test.js` (210 passed)
- `node --check backend/src/services/ai/providerConfigs.js`
- `node --check backend/tests/providers/providerConfigs.test.js`
- live `GET https://api.atlascloud.ai/v1/models` confirmed the default model and its tools, vision, and reasoning capabilities
- one live chat-completions request returned the exact expected response

### DIFF
--- a/backend/src/services/ai/providerConfigs.js
+++ b/backend/src/services/ai/providerConfigs.js
@@ -1297,6 +1297,17 @@ const PROVIDER_CONFIGS = [
 
 export const PROVIDER_TEMPLATES = [
   {
+    key: 'atlascloud',
+    name: 'Atlas Cloud',
+    baseURL: 'https://api.atlascloud.ai/v1',
+    defaultModel: 'qwen/qwen3.5-397b-a17b',
+    supportsTools: true,
+    supportsVision: true,
+    supportsStreaming: true,
+    description:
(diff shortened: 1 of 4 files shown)


agnt-gg:
Thanks for the Atlas Cloud template contribution! I’ve preserved your original commit and added a small maintainer fix so the existing template dropdown renders above the custom-provider dialog instead of behind it.

The additional production change is one z-index prop on that dialog’s selector, with a regression test exercising the real selector and Atlas form population. No global dropdown changes, credentials, or built-in provider behavior were changed.

The current-main candidate passes 4,429 frontend tests and 227 focused provider tests. The remaining GitHub checks are running; once green, I’ll rebase-and-merge this original PR so your contribution is credited and recorded as merged.

— Annie

PR #113, opened 2026-09-08
fix(goals): load discovered tools before the next model call

tryingET:

A goal task or `run_agent` invocation can call `discover_tools({operation: "load", categories: ["artifact_code"]})` and receive “available in your next response”, yet the next model request still contains only the initial tools. The worker then reports that it cannot read/write files even though the tools are registered and usable in chat.

Two wiring gaps cause this:
- Both `LlmExecutionService` tool loops leave `_requestedToolCategories` unconsumed.
- `TaskOrchestrator.executeTaskViaAgentChat` discards the context returned by `buildAgentRuntime`, including the already-resolved discovery ceiling and prompt state.

This PR fixes that runtime parity issue. It does not add tools to anyone's permissions, alter the authorization model, or change credentials/authentication.

## How does it solve it?

- Preserve `buildAgentRuntime().context` in the goal / `run_agent` caller, with trusted caller identity retaining precedence.
- Share one pending-category consumer between the streaming and non-streaming execution loops, before the next context-budgeting/model call.
- Use existing `getAvailableToolSchemas({userId, asyncEnabled})`, `getToolsForCategories`, and `stripProviderIncompatibleTools`; no duplicate catalog or dependencies.
- Apply the existing resolved `_toolCeiling` (or existing `enabledTools` fallback). Empty ceilings stay empty. Provider-incompatible tools are not reintroduced by discovery.
- Append/deduplicate without replacing the original schemas or reordering their prefix; keep `context.toolSchemas` and loaded/order bookkeeping aligned.
- Clone mutable discovery sets/order arrays per invocation so concurrent executions cannot share pending/loaded state through a reused parent context.
- Reject registry-loading errors explicitly rather than silently making another request with stale schemas. Preserve goal cancellation responsiveness during the new async lookup.

Only two production files change. Current-main computer image/ordering behavior is preserved. No schema migration, provider change, frontend change, service startup, or permission-policy change.

### Review boundaries

The category selection, permission policy and central tool dispatcher are unchanged. This patch propagates/consumes their existing contract; it is not a new authorization or action gate. It also does not attempt a general tool-loop refactor, change existing streaming cancellation semantics, or fix unrelated goal grading/completion behavior.

## How did you verify it?

Based on current upstream `main` at `3090bf66e69f8e1c7ac538beb0493aeb8bca45a5`; no dependency on local-only commits.

### Regression and integration evidence

```text
New tests: 29 passed
Unpatched upstream + the same tests: 23 failed, 6 passed

Full backend suite: 374 files, 5,903 passed, 0 failed
Full frontend suite: 257 files, 4,412 passed, 0 failed
Local running-checkout source regression suite: 112 passed

git diff --check: passed
```

New tests cover both loops, real next-call schema capture, registry errors, same-round/repeated loads, duplicate/order preservation, empty/restricted/fallback ceilings, provider-incompatible names/malformed schemas, user-scoped installed tools, async-disabled settings, concurrency isolation, cancellation during discovery, and caller identity/context/ledger propagation.

The integration test uses the **real `discover_tools` handler, registry/category resolution, central dispatcher and file-reading code tool**, reading a disposable temporary fixture. Only the model/client and ledger are replaced; no live LLM inference or production DB is used. Separate unit tests inject registry/tool failures and user-specific catalogs deterministically.

Three deliberate mutations were tested and restored:
- Remove ceiling filtering: **6 regression failures**.
- Remove provider filtering: **2 regression failures**.
- Share discovery sets instead of cloning: **2 regression failures**.

### Reproduction commands / environment notes

Tested on Linux with Node 26.8.1, using the existing dependency installs (no package/lock changes). Run from the repo root:

```sh
TMPDIR=/tmp AGNT_DISABLE_EXTERNAL_POLLING=true npm test
NODE_OPTIONS=--no-experimental-webstorage TMPDIR=/tmp \
  AGNT_DISABLE_EXTERNAL_POLLING=true npm --prefix frontend test
```

Use a short, non-hidden temporary directory appropriate for your OS; `/tmp` is illustrative. Actual successful runs used a short non-hidden workspace directory. The first backend attempt reproduced 13 `portableBundle` hidden-temp-path failures on both patched and unpatched upstream; a long temp path then hit the existing Brave startup test. Neither suite was changed. With a short non-hidden temp directory, all backend tests passed. Node 26's experimental global Web Storage interfered with JSDOM in the first frontend run; disabling that Node global made all frontend tests pass, without source changes.

The regression-only command is:

```sh
npx vitest run \
  backend/src/services/ai/LlmExecutionService.discovery.test.js \
  backend/src/services/ai/LlmExecutionService.discoveryIntegration.test.js \
  backend/src/services/goal/TaskOrchestrator.runtimeContext.test.js
```

### Not claimed

This is deterministic handler/loop/real-tool integration coverage, not proof from a live model or restarted production goal worker. The local source patch was applied without a service restart; a fresh-process/live-goal smoke remains a deployment check. No changes to inference services were performed. Passing tests are evidence, not a guarantee of defect-free code.

## Checklist

- [x] No auth, session, credential or permission-policy changes; existing runtime ceiling is preserved.
- [x] Non-security runtime parity bug; no new privileges or dispatcher behavior.
- [x] Backend suite passes with the documented test environment.
- [x] Frontend suite passes with the documented Node 26 setting.
- [x] New behavior has tests, observed failing on unpatched upstream.
- [x] Internal behavior documented inline; no public API/configuration change.
- [x] Commit subject is fewer than 72 characters.

### DIFF
--- a/backend/src/services/ai/LlmExecutionService.discovery.test.js
+++ b/backend/src/services/ai/LlmExecutionService.discovery.test.js
@@ -1,0 +1,198 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+
+vi.mock('./LlmService.js', () => ({ createLlmClient: vi.fn(async () => ({})) }));
+vi.mock('../orchestrator/llmAdapters.js', () => ({ createLlmAdapter: vi.fn() }));
+vi.mock('../orchestrator/tools.js', () => ({ executeTool: vi.fn(), getAvailableToolSchemas: vi.fn() }));
+vi.mock('../../utils/contextManager.js', () => ({
+  manageContext: vi.fn((messages) => ({ messages, managedTokens: 1 })),
+  estimateToolTokens: vi.fn(() => 1),
+}));
+vi.mock('../execution/LedgerRecorder.js', () => ({ recordLlmCall: vi.fn(async () => {}) }));
+
+import service from './LlmExecutionService.js';
+import { createLlmAdapter } from '../orchestrator/llmAdapters.js';
+import { executeTool, getAvailableToolSchemas } from '../orchestrator/tools.js';
+import { manageContext } from '../../utils/contextManager.js';
+// Category selection and provider compatibility are REAL modules in this suite.
+const schema = (name, description = name) => ({ type: 'function', function: { name, description, parameters: { type: 'object', properties: {} } } });
+const call = (name, args = {}) => ({ id: `${name}-${JSON.stringify(args)}`, type: 'function', function: { name, arguments: JSON.stringify(args) } });
+const reply = (...toolCalls) => ({ responseMessage: { role: 'assistant', content: toolCalls.length ? null : 'Done', ...(toolCalls.length ? { tool_calls: toolCalls } : {}) }, toolCalls });
+const names = (schemas) => schemas.map((s) => s.function.name);
+let tmp;
+
+
+function adapterFor(responses) {
+  const requests = [];
+  const run = vi.fn(async (messages, schemas) => {
+    // Snapshot: arrays are intentionally extended in place between rounds.
+    requests.push({ schemas: structuredClone(schemas), messages: structuredClone(messages) });
+    return responses.shift() || reply();
+  });
+  const adapter = { call: run, callStream: run, formatToolResults: (results) => results };
+  createLlmAdapter.mockResolvedValue(adapter);
+  return { requests, run };
+}
+function execute(mode, overrides = {}) {
+  const config = { provider: 'openai', model: 'test', userId: 'owner-a', messages: [{ role: 'user', content: 'Discover then read.' }], toolSchemas: [schema('discover_tools')], maxToolRounds: 4, ...overrides };
+  return mode === 'stream' ? service.executeWithToolsStreaming(config, vi.fn()) : service.executeWithTools(config);
+}
+
+beforeEach(async () => {
+  vi.clearAllMocks();
+  service.cacheEnabled = false;
+  service.cache.clear();
+  tmp = await fs.mkdtemp(path.join(os.tmpdir(), 'agnt-discovery-'));
+  await fs.writeFile(path.join(tmp, 'fixture.txt'), 'ACTUAL_FIXTURE_BYTES');
+  getAvailableToolSchemas.mockResolvedValue([schema('read_file'), schema('write_file'), schema('web_search')]);
+  executeTool.mockImplementation(async (name, args, _auth, context) => {
+    if (name === 'discover_tools') {
+      context._requestedToolCategories ??= new Set();
+      for (const category of args.categories || ['artifact_code']) context._requestedToolCategories.add(category);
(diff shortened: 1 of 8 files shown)


tryingET:
## Prompt-cache behavior clarification

This is a correctness/runtime-parity fix, **not a cache-neutral discovery implementation**. “Stable prefix” in the implementation means retaining the original tool definitions/order, not preserving the entire provider prompt prefix after tools are added.

A local capture through the actual Anthropic adapter with a recording SDK confirmed:
- Before discovery: tools=[discover_tools], cache_control on index0.
- After discovery: tools=[discover_tools,read_file], cache_control on index1; original definitions (excluding cache markers) and system text remain unchanged.
- Repeating that expanded request produced identical serialization.

Anthropic documents a tools -> system -> messages prefix hierarchy, with ordinary tool-definition changes invalidating the tool/system/message caches. Thus an actual schema addition can require a cache rewrite even though the old tools remain ordered and the system string is unchanged. Repeated loads adding no schemas avoid further tool-section changes. This does not flush caches or disable subsequent hits; reuse still depends on matching prefixes, TTL, thresholds and provider behavior. The pre-fix goal loop kept the tool array unchanged only because it failed to load the requested tools.

Provider-native deferred/tool-search mechanisms can preserve the prefix by introducing definitions through conversation items instead of growing the leading tool list. That is a separate capability-specific integration, not part of this PR, and must not be assumed for an Anthropic-compatible local endpoint. Another option for predictable tasks is predeclaring a bounded permitted tool set, with a larger initial context cost.

AGNT's existing local non-stream response-cache key/TTL/read-write conditions are unchanged; tool-bearing invocations already bypass response reuse. Provider cache-hit tokens, cache writes and cost were **not measured** in this PR's fake-model tests. No claim of cache-hit-rate nonregression is made.

Reference: https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-use-with-prompt-caching

agnt-gg:
Thanks, @tryingET. I reproduced the missing-tool behavior on current main in both streaming and non-streaming execution, then confirmed your unchanged patch fixes both using the real discovery registry and file reader with a synthetic fixture and mocked model.

Local verification passed 73 focused checks, 5,904 backend tests (1 skipped), 4,429 frontend tests, and the production frontend build. One initial integration run exceeded the default five-second timeout during registry initialization; the local rerun used a 30-second limit. GitHub’s normal checks are now running as the final gate.

No corrective production changes were needed. Once the checks pass, I’ll rebase-and-merge this original PR, preserving your attribution and merged status.

— Annie

agnt-gg:
The browser gate exposed an existing Agents-screen timing bug: when navigation overlaps the startup fetch, the screen snapshots an empty list and fails to follow the eventual store update. I reproduced the exact missing-agent assertion with a deliberately held response; the two-line screen fix makes that regression pass without extending the assertion timeout or removing checks.

I also corrected the cross-client browser harness to reuse the application's shared alias map, removing its unresolved @llm import. These are in one maintainer commit; your original discovery commit is preserved.

Locally, all 25 tagged browser tests and 4,429 frontend tests pass. The candidate is now updated against current main and GitHub checks are running again. No merge until those gates are green.

— Annie

PR #143, opened 2026-09-12
fix(goals): show NEEDS REVIEW and clear stale chat progress

tryingET:
## Problem
A goal with all tasks completed stopped at `needs_review` after evaluation failed, but its embedded chat widget kept showing EVALUATING/RUNNING. Reload cleared transient state and instead showed STARTING. Neither described the authoritative goal state.

## Change
Frontend only, three files:
- GoalProgressWidget explicitly renders NEEDS REVIEW and paused/stopped states. Authoritative review/failure/stopped state overrides stale live data and completion fallback; task counts remain visible without implying acceptance.
- Goals store clears obsolete live-iteration state on inactive/review updates and list hydration. Legitimate executing/evaluating progress remains intact.
- Status hydration updates the goal even if the task list is empty, allowing a reloaded chat widget to recover its authoritative state.

No goal retry, approval, execution, migration, backend restart, or backend behavior changes. No dependency on the open goal/recovery/telemetry PRs. Based on current upstream main 3a21c8e7.

## RED → GREEN
Repository-discovered Given/When/Then tests, mounted Vue component + Vuex:
- Against upstream: **9 failed / 1 passed**.
- Same assertions with patch: **10/10 passed**.
- Includes live evaluating→needs_review transition, reload hydration, empty-task status responses, stale-state precedence, task-count preservation, and a positive executing-state control.
- `npm --prefix frontend test -- --run src/store/features/goals.reviewState.spec.js`
- `npm --prefix frontend run build` passed (bundle-size advisory warnings only).
- `git diff --check` passed.

Tests stub browser storage/API configuration to isolate the suite from Node26 localStorage behavior. Fetch fixtures verify GET-only status hydration, not an acceptance or retry action. Full frontend suite not run locally; CI results pending.

## Scope / rollout
The equivalent patch was applied and built in the local daily checkout under explicit UI-fix authorization before publication; its served bundle was verified. User-visible refreshed screen acceptance has not been claimed. This PR was separately reproduced/tested/built in a clean worktree; unrelated local changes excluded. No backend restart or goal state mutation occurred.

This does NOT fix the separate evaluation-report HTTP500, repeated grading of unchanged tasks, event-generation fencing, or add a Discuss-with-Annie action. It corrects the existing widget/store display path. Build/deploy the frontend and refresh clients to activate. Roll back the frontend revision without modifying goal records.

### DIFF
--- a/frontend/src/store/features/goals.js
+++ b/frontend/src/store/features/goals.js
@@ -92,6 +92,13 @@ const getters = {
   getGoalTaskProgress: (state) => (goalId) => state.goalTaskProgress[goalId] || null,
 };
 
+function clearInactiveProgress(state, goal) {
+  if (['needs_review','validated','completed','failed','error','stopped','paused','planning','pending'].includes(goal.status)) {
+    const { [goal.id]: obsolete, ...rest } = state.liveIteration;
+    state.liveIteration = rest;
+  }
+}
+
 const mutations = {
   SET_LOADING(state, loading) {
     state.isLoading = loading;
@@ -107,6 +114,7 @@ const mutations = {
 
   SET_GOALS(state, goals) {
     state.goals = goals;
+    goals.forEach(goal => clearInactiveProgress(state, goal));
   },
 
   SET_GOALS_SUMMARY(state, goals) {
@@ -126,7 +134,10 @@ const mutations = {
     const index = state.goals.findIndex((goal) => goal.id === updatedGoal.id);
     if (index !== -1) {
       state.goals.splice(index, 1, { ...state.goals[index], ...updatedGoal });
+    } else if (updatedGoal.status) {
+      state.goals.push(updatedGoal);
     }
+    clearInactiveProgress(state, updatedGoal);
   },
 
   REMOVE_GOAL(state, goalId) {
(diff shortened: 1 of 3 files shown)


agnt-gg:
Merged — thank you. Reproduced first against current `main` before touching anything: your spec ran **9 failed / 1 passed** on the unfixed tree (`RUNNING` where `NEEDS REVIEW` belonged, stale `evaluating` phase surviving terminal updates), then **10 / 10** with the fix. Full frontend suite green at 4,605 with the +10 being exactly your spec, production build green.

`needs_review` is a terminal state on the backend — `TaskOrchestrator` drops the goal from `runningGoals` when it sets it — so shadowing the live payload there is the right call, and the same logic holds for `paused`/`stopped` since nothing runs while a goal sits in them.

Two non-blocking notes for the record, neither needed a change:

- `UPDATE_GOAL` now pushes a stub row when the goal isn't in the list yet. Right for the reload case this fixes; the next `SET_GOALS` replaces it, so the only cost is a brief partial row on the Goals screen.
- `planning` is treated as inactive for live-iteration clearing while the `activeGoals` getter counts it as active. Couldn't find a path that gives a planning goal a live iteration, so it's moot today — just flagging the asymmetry.

I rebased your branch onto current `main` before merging so CI would exercise exactly what landed. Your commit, message, and authorship are untouched.

— Annie

PR #146, opened 2026-09-12
fix(workflows): stop timers, prevent queue spin, guard terminal rows

tryingET:
### Defects and fixes

**F1: Timers outlive workflow stop**
`trigger-timer.js` — The fire-on-start `setTimeout` handle was never registered in `engine.timerIntervals`, so `stopWorkflowListeners()` couldn't disarm it. A stopped workflow with a fire-on-start timer would still execute on the next interval.

**F1: Stopped workflows can spin on a pending queue**
`WorkflowEngine.js` — `_handleTriggerQueue()` skipped draining when `stopRequested` was true but rescheduled itself via `setImmediate` whenever the queue was nonempty — an infinite loop. `stopWorkflowListeners()` now clears the trigger queue; `processWorkflowTrigger()` rejects triggers arriving after stop (covering both the queue path and the synchronous `waitForCompletion` path); `_handleTriggerQueue()` early-returns when stopped.

**F2: Stop/deactivate race overwrites terminal status**
`ExecutionModel.js` — `update()` had no terminal guard, so a deactivate-mediating `'stopped'` write could flip a `'completed'` row to `'stopped'`. Now uses a single-statement `WHERE ... AND status NOT IN ('completed','error','stopped','insufficient-credits')` guard. Same-status terminal writes are also refused (original evidence preserved). No read-then-write race window.

**F3: Late stop writes `undefined` as the stop reason**
`WorkflowEngine.js` — External stops write `undefined` as the log/status argument. Both stop blocks now write a canonical `'stopped'` status via a shared `_finalizeExecution` helper (reason stays in the execution log, not the status column).

**F4: Workflow detail doesn't return the authoritative status**
`WorkflowService.js` — `getWorkflowById()` returned no `status` field at all while every list view returned the authoritative column. Now reads the same `status` column.

**F5: Missing custom-tool definition produces vacuous success**
`NodeExecutor.js` — When a saved custom-tool type wasn't found in the registry, the node warned and fell through to a default AI call (or executed `undefined` as code), producing a "completed" node that did nothing the workflow intended. Now fails with an explicit error naming the missing type before any execution.

### Verification
- **118/118 existing workflow/service tests pass** on upstream main with the patches applied
- **RED→GREEN**: 10 tests fail on unmodified source, 13 pass on patched source (3 are behavioral pins that must keep passing)
- **39/39 repo gates pass** (startStop, failurePersistence, failureReceipt, shapeGuard)
- **Live dogfooded in isolated DEV instance**: 9/9 live checks — inline execution, late stop preserving completion, list/detail parity, missing-definition error, in-flight stop recording canonical `stopped`, no timer restart across 62 seconds (> one full interval)

### Relationship to other PRs
Self-contained. Based on upstream `main` (`3a21c8e7`). Fixes different defects than #138 (failure receipts) and #139 (agent outcomes). No file overlap with those PRs.

### Not included
No frontend changes. No schema migration. No dependency changes.

### DIFF
--- a/backend/src/models/ExecutionModel.js
+++ b/backend/src/models/ExecutionModel.js
@@ -66,8 +66,14 @@ class ExecutionModel {
     return new Promise((resolve, reject) => {
       const safeStatus = status || 'stopped';
 
+      // F2/F3: terminal rows finalize once. Once an execution reached a
+      // terminal status, a different-status write (e.g. a late deactivate
+      // stamping 'stopped' over 'completed') matches zero rows and changes
+      // nothing; same-status writes are also refused, and non-terminal rows
+      // (started/running) transition freely. Single-statement guard: no
+      // read-then-write race window.
       db.run(
-        'UPDATE workflow_executions SET status = ?, log = ?, end_time = ?, credits_used = ? WHERE id = ?',
+        'UPDATE workflow_executions SET status = ?, log = ?, end_time = ?, credits_used = ? WHERE id = ? AND status NOT IN (\'completed\', \'error\', \'stopped\', \'insufficient-credits\')',
         [safeStatus, log, new Date().toISOString(), creditsUsed, id],
         function (err) {
           if (err) reject(err);
--- a/backend/src/services/WorkflowService.js
+++ b/backend/src/services/WorkflowService.js
@@ -271,6 +271,12 @@ class WorkflowService {
         res.json({
           id: workflow.id,
           user_id: workflow.user_id,
+          // F4: list/detail share the authoritative status. The workflows.status
+          // column is what every list endpoint returns and what the engine
+          // writes; the workflow JSON never carries one (saveWorkflow strips
+          // it), so without this field the detail response has no status at
+          // all and can never agree with the list.
+          status: workflow.status,
           workflow: JSON.parse(workflow.workflow_data),
           created_at: workflow.created_at,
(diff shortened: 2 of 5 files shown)


agnt-gg:
Merged — thank you for this. All five defects reproduced exactly as described against current `main` before I touched anything, and the fixes were correct for every one of them. The F2 case was the most striking: a real `completed` row flipped to `stopped` with its credits overwritten, which is precisely the evidence-destroying race you called out.

One follow-up landed directly behind your commit (`3d0792d9`), so a note on what and why:

**The late-trigger guard was keyed on `!isListening` as well as `stopRequested`.** That also rejected an engine that had *never* armed listeners — which is what a sub-workflow engine and a direct programmatic caller both look like on their first run. Keying it on `stopRequested` alone covers every late-trigger case (it's the only thing `stopWorkflowListeners()` has to enforce) without refusing a legitimate first run.

Two smaller things in the same commit: the per-engine `finalizedExecutions` set is now bounded (a listening engine is long-lived and added one id per run forever — the SQL guard is authoritative, the set is just a fast path), and 30 regression tests now pin all five defects plus the never-armed boundary, so they can't silently come back.

I rebased your branch onto current `main` before merging so CI would exercise exactly what landed — `NodeExecutor` had moved underneath the original base. Your commit (`bf2dbc78`), message, and authorship are untouched.

— Annie

PR #147, opened 2026-09-18
feat(connectors): add TypeSafe as an API-key connector tile

rimusz:
## What

Adds **TypeSafe AI** as an API-key connector tile on **Connectors → API/OAuth**.

TypeSafe's Jev is a typed-decision API — it returns schema-locked answers with calibrated probabilities, not chat completions. There is currently no working way to store a key for it:

- the remote catalogue at `api.agnt.gg/auth/providers` does not list `typesafe`, and `POST`ing a new provider there needs a live session;
- a hand-filled **custom provider** works, but asks the user to know the id, connection type and endpoint;
- **Vault** looks like the right place and is not — it writes `user_secrets` in `localStorage`, which `AuthManager` never reads.

## How

A small catalogue of API-key connectors (`connectorCatalog.js`, one entry today) merged into the provider list client-side, so the tile appears whether or not the remote catalogue ever carries the row.

**1. Merged on both fetch paths.** `appAuth.fetchAllProviders` injects the catalogue on the success path *and* the offline fallback. The fallback matters most — it is what a user sees when the catalogue call fails, which is exactly when a client-side row is the only way to reach the provider.

**2. Kept out of the model picker.** `connectableAiProviders()` drops rows marked `connectorOnly`. Without this, a provider with no chat completions endpoint is offered as a chat model and fails at first use. Its category stays `AI` so it remains findable under Connectors.

**3. Resolvable by the backend.** `AuthDispatcher.getAuthEntry()` falls back to the connector catalogue, so `POST /api/providers/:id/auth/connect` resolves the id and stores the key. That route already encrypts into `api_keys` for remote-type providers, so no new storage path was added. `TYPESAFE_API_KEY` is honoured from the environment like any other API-key provider.

**4. Saved to the right store, from every screen.** Connector keys go to that local route rather than the remote key store, which has no provider to attach them to.

## The part that was not one line

Point 4 looked like a single branch in `Connectors.vue`. It was three.

This app has several copy-pasted api-key save paths. Patching one left the others posting connector keys to a store that cannot hold them — so the same key saved from the Connectors screen worked, and from the Tools screen silently did not:

| Path | Reaches connector rows? | |
|---|---|---|
| `Connectors.vue` | yes — maps `allProviders` directly | fixed |
| `composables/useProviderConnection.js` | yes — Tools, ToolsPanel, WorkflowForge panel | fixed |
| `ChatPanel/IntegrationHealth.vue` | yes — reads `allProviders` from the store | fixed |
| `OnboardingModal.vue` | no — renders via ProviderLanes | exempt |
| `Chat/ProviderSetup.vue` | no — renders via ProviderLanes | exempt |
| `Settings/_OauthManager.vue` | no — orphan, fetches its own list | exempt |

The save is now one shared function, `services/connectorApiKey.js`, called from the three reachable paths.

The two `ProviderLanes` screens deliberately do **not** get the branch. It would be unreachable — `providerLanes()` filters `connectorOnly` rows out upstream — and `store/app/providerGridParity.spec.js` exists specifically to keep provider logic out of those two files. I had added it to both before that guard told me otherwise; reverting was the right call.

`__guards__/connectorKeyStoreContract.spec.js` discovers save paths from source and holds every one of them to the rule, so a fourth path added tomorrow is covered. Both exemptions are **asserted rather than assumed**: the orphan is checked to still be unreferenced, and the lane screens are checked to still render through `ProviderLanes`.

## Open question for review: duplication vs. `@llm`

The catalogue is **duplicated** — `frontend/src/store/auth/connectorCatalog.js` and `backend/src/services/auth/connectorCatalog.js`.

That deviates from where this repo is going. `build/aliases.js` defines `@llm` → `backend/src/services/ai/descriptor` precisely so one fact does not live in two places, after hand-mirrored provider predicates drifted, and `descriptor.purity.test.js` enforces the shared modules stay isomorphic.

I kept two copies because the row carries auth-only fields (`authScheme`, the `getConnectorProvider` lookup) and sits naturally in `services/auth/`, not in the AI descriptor tree — so sharing it means moving it under `descriptor/` or adding a second alias, both of which touch build config.

The duplication is fully guarded: the frontend spec **imports** the backend module and compares rows as objects, field by field, deriving the field list from the rows themselves — so a field added later is covered without anyone remembering the test exists. Comparing objects rather than grepping source also catches a changed array member, a renamed field, or the curly apostrophe in `instructions` surviving one file's encoding and not the other's.

**Happy to move it under `@llm` instead if that is preferred** — say so and I will rather than defend the copy.

## Verification

Run on `feat/typesafe-connector-tile` @ `e2434c1`, one commit on top of `ced0a2a`. 17 files, +842/−5.

**Added tests — 35**

| Tests | File | Covers |
|---|---|---|
| 4 | `AuthDispatcher.connectorCatalog.test.js` | id resolves while `getProviderConfig` stays falsy; `connect-apikey` capability; env mapping; unknown ids |
| 8 | `store/auth/connectorCatalog.spec.js` | merge behaviour, plus four drift guards against the backend copy |
| 9 | `services/connectorApiKey.spec.js` | the shared save: local route used, remote never touched, `success:false` throws, malformed response throws, transport error propagates |
| 4 | `useProviderConnection.connectorKey.spec.js` | the real toggle path saves locally; an ordinary provider still posts remotely; a failure reports rather than claims success; no fallback to the remote store |
| 8 | `__guards__/connectorKeyStoreContract.spec.js` | every discovered save path calls the guard **before** posting; both exemptions still hold; the matchers themselves |
| 2 | existing specs | absent from `connectableAiProviders()`; injected on the remote-success path |

**Suites** — frontend 4622 pass / 282 files · `backend/src/services/auth/` 366 pass · backend full 6176 pass / 410 files · `vite build` clean.

**Mutation-tested — 23 mutations, 23 killed, 0 survived.**

- *Behaviour (4):* removing the `connectorOnly` filter, either `appAuth` merge, the `AuthDispatcher` fallback.
- *Catalogue drift (13):* each of the seven shared fields altered in the backend copy; `authScheme` altered; a field removed; a field added to one copy only (each direction); the row removed; a second row added.
- *Key save (6):* the connector branch deleted from each of the three save paths; the shared save made to swallow `success:false`, to post remotely, and to always report "not a connector".

Two mutations initially **survived**, and fixing that mattered: the guard tested `code.includes('usesLocalKeyStore')`, which is still satisfied by the `import` line after the branch itself is deleted — so it passed on a reverted fix. It now requires an actual call, with a test pinning that an import alone does not vouch for a file. Failure is detected by vitest's exit code; its summary line carries ANSI escapes between `Tests` and the count, so grepping that under-reports.

**Still only structurally covered.** `saveApiKey` in `Connectors.vue` and `IntegrationHealth.vue` are private closures inside large SFCs with no unit spec, so their call sites are held by the source-level guard rather than by a behavioural test. The guard does kill the mutation that deletes either branch — but it proves the call is *present*, not that it *works*. The logic they call is unit-tested, and the equivalent path through the composable is tested behaviourally.

**Pre-existing failures.** Three backend tests fail on this machine with *and* without this branch: `browserActDriver.live` and `browserFallbackSurface` ×2 (no Brave/Chromium installed) and `goal/controlledLifecycle.bdd` (`DB escaped disposable directory`). Confirmed on a pristine `origin/main` worktree.

## Scope

Additive. No existing provider changes behaviour: the `appAuth` edits wrap an existing commit payload, the `aiProvider` filter only rejects ids present in the new catalogue, `getAuthEntry` is unchanged for every id `providerConfigs` already resolves, and each save path's new branch is entered only for a connector row.

**Nothing here calls TypeSafe.** No request is made to `api.typesafe.ai`; this only lets a user store a key that something else can later read.

**The tool gate that motivated this is not included.** That work loads policy from a user-writable directory outside the repo and is not ready for review; it will be a separate PR, or none.

Also deliberately excluded: a "Quick add" preset chip on the custom-provider form. The catalogue merge makes it redundant.


agnt-gg:
Thanks for putting this together! **Adding your TypeSafe/JEV provider and API key is already supported through AGNT’s existing interface—no plugin or core changes are needed.**

In your instance, go to **Connectors → Add New Integration**, register the provider with an API-key connection type, then connect your key through the existing credential flow. The provider is account-scoped; it does not need to be globally registered for every AGNT user.

We’ve discussed keeping integrations out of core auth, and this PR still crosses that boundary: it modifies `AuthDispatcher`, introduces frontend/backend connector catalogs, injects them into `appAuth`, and adds special credential-saving behavior. Please use the existing registration and credential system instead.

The local definitions for CLI-backed providers are a specific exception—not the pattern for adding regular API-key providers.

**Reference:** [Auth-provider API documentation](https://github.com/agnt-gg/agnt/blob/main/docs/_API-DOCUMENTATION.md#remote-auth-routes)

- Provider registration: `POST https://api.agnt.gg/auth/providers`
- Existing key-storage endpoint: `POST https://api.agnt.gg/auth/apikeys/:providerId`

The UI handles the existing credential flow, so there’s no need to implement your own key-storage request or auth path. Also, `/api/custom-providers` is for OpenAI-compatible model providers, not generic auth integrations like JEV.

**I’m closing this PR in favor of configuring the provider through the existing system.** No plugin, fork, or hardcoded catalog is required! If that flow is failing on your instance, please share the specific reproduction so we can address the actual issue without adding a parallel mechanism.

— Annie ✨

